PR #80, opened 2018-09-18
swapping fragments in root, back stack manipulation

sashei:


app/src/main/java/mozilla/lockbox/presenter/IntentPresenter.kt
```diff
@@ -1,20 +1,59 @@
...
+
+    val welcome: WelcomeFragment by lazy { WelcomeFragment() }
+    val login: FxALoginFragment by lazy { FxALoginFragment() }
 
     init {
         routeStore.routes.subscribe { a -> route(a) }
     }
 
+    fun onViewReady() {
+        replaceFragment(this.welcome, false)
```
jhugman: This is where we might need to worry about persisting/restoring the current route.

PR #86, opened 2018-09-19
make sure to not leak intentpresenter

sashei:
disposal example


app/src/main/java/mozilla/lockbox/presenter/IntentPresenter.kt
```diff
@@ -17,8 +18,10 @@ class IntentPresenter(private val activity: AppCompatActivity, routeStore: Route
     val welcome: WelcomeFragment by lazy { WelcomeFragment() }
     val login: FxALoginFragment by lazy { FxALoginFragment() }
 
+    val compositeDisposabe = CompositeDisposable()
+
     init {
-        routeStore.routes.subscribe { a -> route(a) }
+        compositeDisposabe.add(routeStore.routes.subscribe { a -> route(a) })
```
linuxwolf: 🤔 is this something we can add an extension to Disposable to chain things a tad cleaner?

PR #97, opened 2018-09-20
Test all routes with a navigator

jhugman:
This PR introduces a navigator, a reusable object to make writing UI testing easier.

This is in service of testing the `RoutePresenter`.


devinreams:
>>Lint task failed, error: exit status 1

>w: /bitrise/src/app/src/main/java/mozilla/lockbox/presenter/SettingPresenter.kt: (9, 24): Parameter 'view' is never used

This will be the case and keep erroring for a little bit... we can suppress it for now until it comes in.

>/bitrise/src/app/src/[REDACTED]Test/java/mozilla/lockbox/Navigator.kt:17:1: Needless blank line(s)

You're likely aware already but ktlint errors like this one can almost always be resolved by running `./gradlew ktlintFormat` 🏃

PR #106, opened 2018-09-24
Update Android Studio and kotlin and gradle

devinreams:
I don't think this will build successfully, there should be errors.


devinreams:
OK this should be working.

@sashei @jhugman @linuxwolf Do you want to / are you ready to update to Android Studio 3.2 and these other things?

I dropped updating android-components here because of additional logins mocking needed...

sashei:
@devinreams ready!!

jhugman:
I'm ready here.

devinreams:
Thanks, here we go to the latest and greatest! 🚀

PR #107, opened 2018-09-24
Add Sentry to the project

### LINKED ISSUE #105: Add Sentry for error reporting
https://docs.sentry.io/clients/java/modules/android/?platform=javascript

devinreams:
Fixes #105 

- [x] add Sentry
- [x] test a failure/exception and make sure it records in Sentry
- [x] add to ProGuard and upload the rules to Sentry
- [x] set DSN as variable

### DIFF
--- a/app/lint-baseline.xml
+++ b/app/lint-baseline.xml
@@ -1,55 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-  ~ This Source Code Form is subject to the terms of the Mozilla Public
-  ~ License, v. 2.0. If a copy of the MPL was not distributed with this
-  ~ file, You can obtain one at http://mozilla.org/MPL/2.0/.
-  -->
-
-<issues format="4" by="lint 3.1.4">
+<issues format="4" by="lint 3.2">
 
     <issue
         id="UnusedAttribute"
         message="Attribute `startY` is only used in API level 24 and higher (current min is 23)">
-        <location
-            file="src/main/res/drawable/ic_nessie.xml"
-            line="40"/>
-    </issue>
-
-    <issue
-        id="UnusedAttribute"
-        message="Attribute `startX` is only used in API level 24 and higher (current min is 23)">
-        <location
-            file="src/main/res/drawable/ic_nessie.xml"
-            line="41"/>
-    </issue>
-
-    <issue
-        id="UnusedAttribute"
-        message="Attribute `endY` is only used in API level 24 and higher (current min is 23)">
-        <location
-            file="src/main/res/drawable/ic_nessie.xml"
-            line="42"/>
-    </issue>
-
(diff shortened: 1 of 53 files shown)


devinreams:
OK I have the test Sentry crashes reporting:

image

https://sentry.prod.mozaws.net/operations/lockbox-android/dashboard/

There is more we can do if we want (set the "environment") but this feels like a good place to start...

sashei:
✨emails✨

devinreams:
OK I added proguard rules and was going down a path of having it uploading rules but realized we may not actually obfuscate our code so don't need it. I'm leaving it out for now.

This is ready for review! ✅

app/src/test/java/mozilla/lockbox/store/DataStoreTest.kt
```diff
@@ -58,7 +58,7 @@ class DataStoreTest : DisposingTest() {
         subject.unlock().subscribe(waiter)
         waiter.assertComplete()
 
-        waiter = createTestObserver<Unit>()
+        waiter = createTestObserver()
```
sashei: nice.

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -6,15 +6,24 @@
...
 import android.app.Application
+import io.sentry.Sentry
+import io.sentry.android.AndroidSentryClientFactory
 import mozilla.components.support.base.log.Log
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.log.sink.AndroidLogSink
 
 val log: Logger = Logger("Lockbox")
 class LockboxApplication : Application() {
+    @SuppressLint("AuthLeak")
```
sashei: is this actually OK?

app/lint-baseline.xml
```diff
@@ -1,1090 +1,1112 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
```
sashei: 😬 everything about this change is a bummer

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
devinreams: Eh.. no, it's not great. I will see about turning it into an environment variable.

app/lint-baseline.xml
devinreams: I'll start chipping away at the baseline next so we see far less of this nonsense, I promise 😁

app/src/test/java/mozilla/lockbox/store/DataStoreTest.kt
devinreams: >nice.

All credit goes to the Android linter ❇️

PR #112, opened 2018-09-27
Item Detail View

### LINKED ISSUE #5: Display item details view
When I need to access an account, I want to see an entry's details to either copy credentials to my clipboard or view the username / password, so that I can easily put those credentials into the app or mobile browser.

## Acceptance criteria
* Should have a details view with all relevant details including title, URL, username, password (not revealed by default) 
* Should be informed about how to edit/delete

### Visual Design

**InVision**

**Zeplin**
https://zpl.io/2EAZQyl

jhugman:
Fixes #5

This PR introduces a new fragment and presenter, with associated tests. 

Since the fragment is parameterized, the `RouteAction` `enum` has been transformed into a set of sealed classes and static `object`s.

As the fragments are re-used, but with different parameters, the `onResume` method is copied from the fragment to the abstract `Presenter`.


app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -33,7 +34,7 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
         return view
     }
 
-    override fun updateItem(item: ItemViewModel) {
+    override fun updateItem(item: ItemDetailViewModel) {
         inputLayoutHostname.isHintAnimationEnabled = false
```
linuxwolf: The toolbar.title should be set here.

PR #114, opened 2018-09-28
Use Sealed Classes instead of Nested Enums in DataStore

linuxwolf:
Changes the `DataStoreAction` and `DataStore.State` from using nested enums to instead be sealed classes.  This better much more closely aligns with the patterns we have in iOS.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/action/DataStoreAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/DataStoreAction.kt
@@ -8,11 +8,9 @@ package mozilla.lockbox.action
 
 import mozilla.lockbox.flux.Action
 
-data class DataStoreAction(val type: Type) : Action {
-    enum class Type {
-        LOCK,
-        UNLOCK,
-        RESET,
-        SYNC
-    }
+sealed class DataStoreAction : Action {
+    object LOCK : DataStoreAction()
+    object UNLOCK : DataStoreAction()
+    object RESET : DataStoreAction()
+    object SYNC : DataStoreAction()
 }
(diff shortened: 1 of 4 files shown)


devinreams:
>on naming, I am conflicted...

>For reference, ktLint prefers us to use CamelCase

...and if it wasn't obvious, ktLint is carrying out the Kotlin Style Guide:
https://android.github.io/kotlin-guides/style.html

In case its helpful, summarized here:

- "Class names are written in PascalCase and are typically nouns or noun phrases"
- "Function names are written in camelCase and are typically verbs or verb phrases"
- "Constant names use UPPER_SNAKE_CASE: all uppercase letters, with words separated by underscores"
- "Non-constant names are written in camelCase."

linuxwolf:
> So, which naming convention should we use? SNAKE_CAPS, CamelCase or droppedCamelCase?
>
> For reference, ktLint prefers us to use CamelCase, which is what I went for in #112.

I used the snake case only because it was in an earlier commit set, and my own habits from previous iterations of Java dev.  More than happy to switch to CamelCase and match current conventions.

> I have a second, much less important thing to be worried about here. Each of these constants appear to be new classes; are there any warnings/war stories/things to worry about, especially or a 65565 methods nature. One for the android meeting.

It might be worth discussing, although the state of Java and Android runtimes has changed quite significantly in the last 5 years.

Using sealed classes here much closer mimics what is in iOS.  Further, we'll eventually need to update credentials with the datastore, which iOS does with a parameterized DataStoreAction.  I understand the concern, but I'd like evidence we're approaching this limit before we break from what looks to be standard Kotlin practice.

jhugman:
> More than happy to switch to CamelCase and match current conventions.

Yes, please do this.

Issue #93, opened 2018-09-19
Resolve and remove baseline lint warnings


devinreams:
Follow up to #89 

141 warnings were added to baseline so we can block/error on builds going forward

But we need to clean up those warnings...

devinreams:
Now that #152 is merged the baseline is much smaller.

A few things remain but I will move this back into the inbox to revisit later on (closer to launch prep).

PR #126, opened 2018-10-03
Remove vector assets for pre and post API 23

devinreams:
connected to #93 to get our vectors vs. bitmaps in the right spots to a) clear up the lint warnings and b) make sure the assets actually (and quickly) load before and after API 23.

looks like the gradients in nessie don't work on 23 and below so may just be easiest to bundle up bitmaps. i'm putting this PR up to get some thoughts in case anyone has them, otherwise I'll keep playing with and testing this out.


devinreams:
I don't think I'm well equipped to solve this. Lots to consider around SVG vs flat image use here.. I'll close for now to remove clutter.

PR #110, opened 2018-09-26
add metrics.md

irrationalagent:
closes #15 
first version of the metrics plan, sure to evolve more. 

the main differences from [the ios version](https://github.com/mozilla-lockbox/lockbox-ios/blob/master/docs/metrics.md) are:

1. discussion of android-specific telemetry libraries,
2. tweaked some of the hypotheses related to analysis, given autofill will be in v1
3. added a short section on the adjust sdk

PR #124, opened 2018-10-03
Copy and reveal credentials

mihainisipeanusv:
Fixes #13 
Fixes #12 

Adding the username and password copy in ItemDetailFragment, also reveal / hide password

-  please let me know is adding a EditText in ItemDetailsView passed to presenter is the correct approach
 - please let me know what is the correct approach to mock a EditText in tests.

PR #128, opened 2018-10-03
TelemetryStore implementation

### LINKED ISSUE #116: Implement initial TelemetryStore
Implement the basics for the `TelemetryStore` and initial actions.  This effectively wraps the `service-telemetry` Android component, and establish its configuration based on requirements from #15.

blocks #14

linuxwolf:
Fixes #116

Implements a Flux store + actions around Telemetry, using `service-telemetry` as the basis.  This work takes the requirements drafting in #110 into account.

**NOTE:**  The enumerations are defined to exactly map to their metrics values rather than adhere, in order to take as much advantage of the system's defaults as we can.  This means they deviate from coding standards.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
+++ b/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
@@ -15,6 +15,7 @@ import mozilla.components.support.base.log.Log
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.log.sink.AndroidLogSink
 import mozilla.lockbox.store.ClipboardStore
+import mozilla.lockbox.store.TelemetryStore
 
 val log: Logger = Logger("Lockbox")
 class LockboxApplication : Application() {
@@ -24,6 +25,9 @@ class LockboxApplication : Application() {
 
         // use context for system service
         ClipboardStore.shared.apply(getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager)
+
+        // hook this context into Telemetry
+        TelemetryStore.shared.applyContext(this)
 
         // Set up Sentry using DSN (client key) from the Project Settings page on Sentry
         val ctx = this.applicationContext
--- a/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
@@ -10,10 +10,10 @@ import mozilla.lockbox.flux.Action
 import org.mozilla.telemetry.event.TelemetryEvent
 
 open class TelemetryAction(
-    val eventMethod: TelemetryEventMethod,
-    val eventObject: TelemetryEventObject,
-    val value: String?,
-    val extras: Map<String, Any>?
+    open val eventMethod: TelemetryEventMethod,
+    open val eventObject: TelemetryEventObject,
+    open val value: String?,
+    open val extras: Map<String, Any>?
 ) : Action {
     open fun createEvent(category: String = "action"): TelemetryEvent {
         val evt = TelemetryEvent.create(
--- a/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
@@ -15,17 +15,17 @@ open class TelemetryAction(
     val value: String?,
     val extras: Map<String, Any>?
 ) : Action {
-  open fun createEvent(category: String = "action"): TelemetryEvent {
-      val evt = TelemetryEvent.create(
-              category,
-              eventMethod.name,
-              eventObject.name,
-              value
-      )
-      extras?.forEach { ex -> evt.extra(ex.key, ex.value.toString()) }
+    open fun createEvent(category: String = "action"): TelemetryEvent {
+        val evt = TelemetryEvent.create(
+                category,
+                eventMethod.name,
+                eventObject.name,
+                value
+        )
+        extras?.forEach { ex -> evt.extra(ex.key, ex.value.toString()) }
 
-      return evt
-  }
+        return evt
+    }
 }
 
 enum class TelemetryEventMethod {
--- a/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
@@ -7,13 +7,26 @@
 package mozilla.lockbox.action
 
 import mozilla.lockbox.flux.Action
+import org.mozilla.telemetry.event.TelemetryEvent
 
 open class TelemetryAction(
     val eventMethod: TelemetryEventMethod,
     val eventObject: TelemetryEventObject,
     val value: String?,
     val extras: Map<String, Any>?
-) : Action
+) : Action {
+  open fun createEvent(category: String = "action"): TelemetryEvent {
+      val evt = TelemetryEvent.create(
+              category,
+              eventMethod.name,
+              eventObject.name,
+              value
+      )
+      extras?.forEach { ex -> evt.extra(ex.key, ex.value.toString()) }
+
+      return evt
+  }
+}
 
 enum class TelemetryEventMethod {
     tap,
--- a/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
@@ -1,0 +1,58 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.action
+
+import mozilla.lockbox.flux.Action
+
+open class TelemetryAction(
+    val eventMethod: TelemetryEventMethod,
+    val eventObject: TelemetryEventObject,
+    val value: String?,
+    val extras: Map<String, Any>?
+) : Action
+
+enum class TelemetryEventMethod {
+    tap,
+    startup,
+    foreground,
+    background,
+    settingChanged,
+    show,
+    canceled,
+    login_selected,
+    autofill_locked,
+    autofill_unlocked,
+    refresh,
+    autofill_clear
+}
+
+enum class TelemetryEventObject {
+    app,
+    entry_list,
+    entry_detail,
+    learn_more,
+    reveal_password,
+    entry_copy_username_button,
+    entry_copy_password_button,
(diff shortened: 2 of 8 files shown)


linuxwolf:
@mozilla-lockbox/mobile-engineering The `service-telemetry` API requires a `Context` when an instance is created.  Herein is one approach to dealing with that, by exposing a `applyContext()` method that should be called in `LockboxApplication.onCreate()`.  However, I'd really like to see if there's a better way to do this.

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -12,13 +12,17 @@ import io.sentry.android.AndroidSentryClientFactory
...
+import mozilla.lockbox.store.TelemetryStore
 
 val log: Logger = Logger("Lockbox")
 class LockboxApplication : Application() {
     override fun onCreate() {
         super.onCreate()
         Log.addSink(AndroidLogSink())
 
+        // hook this context into Telemetry
+        TelemetryStore.shared.applyContext(this)
```
sashei: I think there's a larger question about how to get context in stores where it's required... @mihainisipeanusv is dealing with the same issue in #124 right now. It might be worth having a discussion about this rather than addressing in these two PRs separately.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
```diff
@@ -0,0 +1,71 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.action
+
+import mozilla.lockbox.flux.Action
+import org.mozilla.telemetry.event.TelemetryEvent
+
+open class TelemetryAction(
```
sashei: Is there a reason this is a `class` and not an `interface`? I'd love to see an example of how this will be implemented on individual `Action`s in this PR.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
```diff
@@ -0,0 +1,71 @@
...
+    val eventMethod: TelemetryEventMethod,
+    val eventObject: TelemetryEventObject,
+    val value: String?,
+    val extras: Map<String, Any>?
+) : Action {
+    open fun createEvent(category: String = "action"): TelemetryEvent {
+        val evt = TelemetryEvent.create(
```
sashei: ohh, default implementations are not a thing in Kotlin...

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
```diff
@@ -0,0 +1,71 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.action
+
+import mozilla.lockbox.flux.Action
+import org.mozilla.telemetry.event.TelemetryEvent
+
+open class TelemetryAction(
+    val eventMethod: TelemetryEventMethod,
```
sashei: These will probably all need to be `open` as well so that they can be overridden by implementing `Action`s. Having a sample case would help shake out these problems.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
```diff
@@ -0,0 +1,71 @@
...
+    show,
+    canceled,
+    login_selected,
+    autofill_locked,
+    autofill_unlocked,
+    refresh,
+    autofill_clear
+}
+
+enum class TelemetryEventObject {
```
sashei: 🎉 way to get these all in early! follow-up stories will be supa easy ^.^

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
```diff
@@ -0,0 +1,84 @@
...
+
+open class TelemetryStore(
+    val dispatcher: Dispatcher = Dispatcher.shared,
+    val wrapper: TelemetryWrapper = TelemetryWrapper()
+) {
+    companion object {
+        val shared = TelemetryStore()
+    }
+
+    internal val compositeDisposable = CompositeDisposable()
```
sashei: This looks like it never gets disposed.

app/src/test/java/mozilla/lockbox/DisposingTest.kt
```diff
@@ -16,4 +18,10 @@ open class DisposingTest {
     open fun tearDown() {
         disposer.clear()
     }
+
+    fun <T> createTestObserver(): TestObserver<T> {
```
sashei: nice!

app/src/test/java/mozilla/lockbox/store/TelemetryStoreTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private lateinit var dispatcher: Dispatcher
+    private lateinit var wrapper: FakeTelemetryWrapper
+    private lateinit var subject: TelemetryStore
+
+    @Before
+    fun setUp() {
+        dispatcher = Dispatcher()
```
sashei: This might be more of a nit / implementation detail / does it actually matter I dunno... but in the other `Test`s we just do this initialization inline with the variable declarations. However I can see the utility of initialization in `@Before` if you are explicitly testing initialization behavior.

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.mozilla.telemetry.config.TelemetryConfiguration
+import org.mozilla.telemetry.event.TelemetryEvent
+import org.mozilla.telemetry.net.HttpURLConnectionTelemetryClient
+import org.mozilla.telemetry.ping.TelemetryCorePingBuilder
+import org.mozilla.telemetry.ping.TelemetryMobileEventPingBuilder
+import org.mozilla.telemetry.schedule.jobscheduler.JobSchedulerTelemetryScheduler
+import org.mozilla.telemetry.serialize.JSONPingSerializer
+import org.mozilla.telemetry.storage.FileTelemetryStorage
+
+open class TelemetryWrapper() {
```
sashei: This should be an `interface` with an implementation both in actual code and in the tests for DI.

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
linuxwolf: @sashei (et al @mozilla-lockbox/mobile-engineering ).

That's reasonable, although it looks like we're following the same pattern with slightly different names.

Unless there's objections to the overall approach, I suggest the discussion start with:

1. agree to the pattern @mihainisipeanusv and I seem to have come to already, with the same names
2. encapsulate it into a common base class for these types of Stores (e.g. `ContextualStore` ?)

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
linuxwolf: This was copied almost exactly from iOS.  I don't see a good reason to make this an interface devoid of implementation right now.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
linuxwolf: It's the constructor, so why can't implementing actions apply it then?

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
linuxwolf: Also the follow up stories should already exist.

app/src/test/java/mozilla/lockbox/store/TelemetryStoreTest.kt
linuxwolf: I was copying from DataStore, which does have some initialization tests.  I can move this out for this test.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
linuxwolf: Default implementations on an interface are not a Java thing, so they're not a Kotlin thing.

I take it this is enough reason to leave it as `open class TelemetryEvent` instead of `interface TelemetryEvent`?

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
linuxwolf: This is on the Store, and somewhat copied from `DataStore`.

When should `compositeDisposable.clean() be called?

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
linuxwolf: This feels an awful lot like "intefaces for interfaces sake", which is a Java anti-pattern.

I can make that change, but I'd like a really good reason why it's necessary over having this default implementation with the testing's `FakeTelemetryWrapper` overriding stuff.

app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
linuxwolf: IRL conversation ... make `TelemetryAction` abstract (and vals abstract) so it cannot be used standalone.

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
linuxwolf: IRL ... not clear how important this is right now.  Actions to take:

1) remove `compositeDisposable` usage
2) run Leak Canary on test devices and simulators to see how it shakes out
   * QUESTION: see if it makes sense in bitrise's device testing?

linuxwolf: ok, so I didn't remove `compositeDisposable` because it felt like a better way to deal with a linting error about unchecked return values.

We should investigate a better way to deal with it ... (already a potential carry-over in just about all the stores).

linuxwolf:
Temporarily rebasing on `linuxwolf/fix-build-errors` to get green from CI.  This resulted in a few extra files for now:

* `app/lint-baseline.xml`
* `app/src/main/java/mozilla/lockbox/action/ClipboardAction.kt`
* `app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt`
* `app/src/test/java/mozilla/lockbox/store/ClipboardStoreTest.kt`

app/src/main/java/mozilla/lockbox/store/TelemetryStore.kt
sashei: 👍

PR #129, opened 2018-10-05
Web address launches site/app (entry detail view)

### LINKED ISSUE #39: Web address launches site/app (entry detail view)
## Acceptance criteria
* Should be able to tap the web URL to get to the app or a link in the user selected default browser via device settings
* Native app launch should be prioritized, regardless of browser selection

### Visual Design

**InVision**

**Zeplin**
https://zpl.io/2EAZQyl

mihainisipeanusv:
Fixes #39 

Open hostname in user default browser


app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -16,13 +18,32 @@ import mozilla.lockbox.store.DataStore
...
     private val dataStore: DataStore = DataStore.shared
 ) : Presenter() {
+
+    override fun onViewReady() {
+        this.view.addressLayoutClicks
+                .subscribe {
+                    view.itemId?.let {
+                        dataStore.get(it)
+                                .subscribe {
+                                    Log.d("hello", "${it!!.hostname}, ${it!!.formSubmitURL}")
```
sashei: Is this ready for review yet? This should be marked `WIP` until debugging statements are cleaned up.

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -32,9 +37,18 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
...
         setupBackable(view)
 
+        addressLayout = view!!.inputHostname
         return view
     }
 
+    override val addressLayoutClicks: Observable<Unit>
+        get() = addressLayout!!.clicks()
+
+    override fun openWebsite(hostname: String) {
```
sashei: Implementation note: I would rather have this work as a `RouteAction` that gets passed thru the `Dispatcher`, both for metric-gathering purposes and because we may have other views that need to open outside websites later.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
mihainisipeanusv: yes is ready, I had removed forgotten log entry

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
mihainisipeanusv: external apps opened as a route action now

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -32,9 +35,13 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
         val view = inflater.inflate(R.layout.fragment_item_detail, container, false)
         setupBackable(view)
 
+        addressLayout = view!!.inputHostname
         return view
     }
 
+    override val addressLayoutClicks: Observable<Unit>
+        get() = addressLayout!!.clicks()
```
jhugman: Use 

```kt
get() = view!!.inputHostname.clicks()
```

Also: I don't understand what `addressLayoutClicks` is referring to. Suggest changing to `hostnameClicks`.

app/src/main/java/mozilla/lockbox/action/RouteAction.kt
```diff
@@ -16,4 +16,5 @@ sealed class RouteAction : Action {
     object Back : RouteAction()
     object LockScreen : RouteAction()
     data class ItemDetail(val id: String) : RouteAction()
+    data class OpenApp(val host: String) : RouteAction()
```
jhugman: Naming: Is this `OpenApp` or `OpenWebsite`

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
mihainisipeanusv: fixed

app/src/main/java/mozilla/lockbox/action/RouteAction.kt
mihainisipeanusv: do you mean to have 2 entries one for apps and one for websites ?

nickbrandt:
@mihainisipeanusv this is working well for me when testing on the simulator. Do we want to go ahead and include the font color change in this PR as well (#0060df - Blue 60)? 

https://zpl.io/2EAZQyl

app/src/main/java/mozilla/lockbox/action/RouteAction.kt
linuxwolf: In my opinion, if the method is called `openWebsite()`, then the Action ought to be called `OpenWebsite` instead of `OpenApp`.

mihainisipeanusv: sorry about confusion, i was thinking that at some point this open feature will open websites and installed apps. I had updated action name as suggested

mihainisipeanusv:
> @mihainisipeanusv this is working well for me when testing on the simulator. Do we want to go ahead and include the font color change in this PR as well (#0060df - Blue 60)?
> 
> https://zpl.io/2EAZQyl

updated with color from design

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -12,16 +12,19 @@ import android.view.View
...
 import kotlinx.android.synthetic.main.include_backable.*
-import kotlinx.android.synthetic.main.include_backable.view.*
 import mozilla.lockbox.R
 import mozilla.lockbox.model.ItemDetailViewModel
 import mozilla.lockbox.presenter.ItemDetailPresenter
 import mozilla.lockbox.presenter.ItemDetailView
 
 class ItemDetailFragment : BackableFragment(), ItemDetailView {
     override var itemId: String? = null
+    private var addressLayout: View? = null
```
jhugman: Redundant property.

app/src/test/java/mozilla/lockbox/presenter/ItemDetailPresenterTest.kt
```diff
@@ -23,10 +23,14 @@ class ItemDetailPresenterTest {
...
         override var itemId: String? = null
         var item: ItemDetailViewModel? = null
+        val tapStub:  PublishSubject<Unit> = PublishSubject.create<Unit>()
 
         override fun updateItem(item: ItemDetailViewModel) {
             this.item = item
         }
+
+        override val hostnameClicks: Observable<Unit>
+            get() = tapStub
```
jhugman: The test compiles. Does it test anything? i.e. please keep the tests the testing the new code, not just the old code.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -16,13 +18,32 @@ import mozilla.lockbox.store.DataStore
...
 class ItemDetailPresenter(
     private val view: ItemDetailView,
     private val dispatcher: Dispatcher = Dispatcher.shared,
     private val dataStore: DataStore = DataStore.shared
 ) : Presenter() {
+
+    override fun onViewReady() {
+        this.view.hostnameClicks
+                .subscribe {
+                    view.itemId?.let {
```
jhugman: This is the third time I've seen this construction in this class, where the only things change is the code at the top of the pyramid (in this case L38). Extract a common method

```kt
doWithCredentials { dispatcher.dispatch(RouteAction.OpenWebsite(it.formSubmitURL!!)) }
```

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -16,13 +18,32 @@ import mozilla.lockbox.store.DataStore
...
+                        dataStore.get(it)
+                                .subscribe {
+                                    if (it != null) {
+                                        dispatcher.dispatch(RouteAction.OpenWebsite(it.formSubmitURL!!))
```
jhugman: Please check that this should be the `formSubmitURL`. I have a feeling it should be the `hostname`. cc @linuxwolf ?

linuxwolf: Technically this should be `hostname`.  In practice I don't _think_ `hostname` and `formSubmitURL` are different, but in theory there's very little in Firefox to encourage/enforce that.

mihainisipeanusv: updated with hostname

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
mihainisipeanusv: removed

.idea/codeStyles/Project.xml
```diff
@@ -35,6 +38,115 @@
...
+                  <NAME>.*</NAME>
+                  <XML_NAMESPACE>.*</XML_NAMESPACE>
+                </AND>
+              </match>
+              <order>BY_NAME</order>
+            </rule>
+          </section>
+        </rules>
+      </arrangement>
+    </codeStyleSettings>
```
jhugman: XML formatting setting for Android, as directed by Android Studio.

sashei: Should this be a separate issue per our retro?

app/src/main/java/mozilla/lockbox/extensions/ServerPassword+.kt
```diff
@@ -15,6 +16,10 @@ fun ServerPassword.toViewModel(): ItemViewModel {
     return ItemViewModel(hostname, username, this.id)
 }
 
+fun ServerPassword.toDetailViewModel(): ItemDetailViewModel {
```
sashei: nice!

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -11,77 +11,79 @@ import io.reactivex.Observable
...
+            }
+            .subscribe(view::updateItem)
+            .addTo(compositeDisposable)
+    }
 
-        view.isPasswordVisible = false
+    private fun handleClicks(clicks: Observable<Unit>, withServerPassword: (ServerPassword) -> Unit) {
+        clicks.subscribe {
+            this.credentials?.let { withServerPassword(it) }
+        }.addTo(compositeDisposable)
```
sashei: nit: `.addTo` should be on a separate line, not tied to the end of the `subscribe` block.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -11,77 +11,79 @@ import io.reactivex.Observable
...
+        val itemId = this.itemId ?: return
         dataStore.get(itemId)
-                .map {
-                    ItemDetailViewModel(it.id, titleFromHostname(it.hostname), it.hostname, it.username, it.password)
-                }
-                .subscribe(view::updateItem)
-                .addTo(compositeDisposable)
+            .map {
+                credentials = it
+                it.toDetailViewModel()
```
sashei: does this mutate `it` or is there an unused result in here?

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -11,77 +11,79 @@ import io.reactivex.Observable
...
+            }
+            .subscribe(view::updateItem)
+            .addTo(compositeDisposable)
+    }
 
-        view.isPasswordVisible = false
+    private fun handleClicks(clicks: Observable<Unit>, withServerPassword: (ServerPassword) -> Unit) {
+        clicks.subscribe {
+            this.credentials?.let { withServerPassword(it) }
```
sashei: I wonder if there's an access race condition with accessing `this.credentials?` in this way.. If the dataStore pushes an update to the item while we are handling a click, it may be getting mutated as we access it here.

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -26,19 +26,17 @@ import mozilla.lockbox.presenter.ItemDetailPresenter
...
-        arguments?.let {
-            val args = ItemDetailFragmentArgs.fromBundle(it)
-            itemId = args.itemId
+        val itemId = arguments?.let {
+            ItemDetailFragmentArgs.fromBundle(it)
+                .itemId
         }
 
-        presenter = ItemDetailPresenter(this)
+        presenter = ItemDetailPresenter(this, itemId)
```
sashei: ooo, nice

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -73,6 +74,7 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
         inputUsername.readOnly = true
         inputPassword.readOnly = true
         inputHostname.readOnly = true
+        inputHostname.isClickable = true
```
sashei: is this something that can be configured via XML?

app/src/test/java/mozilla/lockbox/presenter/ItemDetailPresenterTest.kt
```diff
@@ -43,70 +39,64 @@ class ItemDetailPresenterTest {
...
-        val username = "dogs@dogs.com"
-        val id1 = "id1"
-        val password1 = ServerPassword(
+    private val fakeCredentials: List<ServerPassword> by lazy {
+        listOf(
+            ServerPassword(
                 "id0",
                 "https://www.mozilla.org",
-                username,
+                "dogs@dogs.com",
```
sashei: 🐶

app/src/test/java/mozilla/lockbox/presenter/ItemDetailPresenterTest.kt
```diff
@@ -43,70 +39,64 @@ class ItemDetailPresenterTest {
...
-
-        val list = listOf(password1, password2, password3)
+                timePasswordChanged = 0L
+            )
+        )
+    }
 
-        for (exp in list) {
+    @Test
+    fun `sends a detail view model to view on onViewReady`() {
```
sashei: ooo I love the backtick test names...

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -11,77 +11,79 @@ import io.reactivex.Observable
...
+            .map {
+                credentials = it
+                it.toDetailViewModel()
+            }
+            .subscribe(view::updateItem)
+            .addTo(compositeDisposable)
+    }
 
-        view.isPasswordVisible = false
+    private fun handleClicks(clicks: Observable<Unit>, withServerPassword: (ServerPassword) -> Unit) {
```
sashei: I like this method :))

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
linuxwolf: `ServerPassword` is treated as immutable, for the object instance you have won't change from underneath you.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
jhugman: this doesn't mutate `it`, but does save it as a side effect. This isn't the cleanest way of doing things.

sashei: ohh I see.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
sashei: but the `val credentials` that we are accessing here _is_ mutable, not an instance getting passed down from another Observable.

sashei: it gets mutated on ln. 77, which could run at the same time as this method.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -11,77 +11,79 @@ import io.reactivex.Observable
...
         // now set up the data.
-        val itemId = view.itemId ?: return
+        val itemId = this.itemId ?: return
         dataStore.get(itemId)
-                .map {
-                    ItemDetailViewModel(it.id, titleFromHostname(it.hostname), it.hostname, it.username, it.password)
-                }
-                .subscribe(view::updateItem)
-                .addTo(compositeDisposable)
+            .map {
```
sashei: Also, we should make sure to bind this sequence to the main thread so that we don't accidentally update the view from a background thread.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
linuxwolf: ah, now I see.  Thanks.  At least the _value_ won't change, so the problem isn't quite as deep as it could have been.

jhugman: I'm unsure there is a good answer to this. If we consider two schemes:

 1. Cache-on-display: store the `ServerPassword` in `credentials` as we give them to the view to display.
 2. No-cache: each time you use a string (for display, or copying) then get a fresh copy from the database.

For cache-on-display: 

 * If `credentials` is changed between display and user clicking copy, then the new string is both displayed and copied. 
 * If  database copy of the credential is changed between the click being detected and the copy taking place, then it will have to wait until the copy has finished, as the click, display and mutation are all on the UI thread.

For the no-cache scheme, the copy may end up copying a string newer string than that is being displayed. The user only finds this out when the user gets to the other app. 

HOWEVER: both of these scenarios are somewhat tiny considerations in comparison to the S L O W speed of the user and app switching: 

1. the user navigates to the detail screen
2. the credentials on the screen are updated
3. the user copies the old username
4. the user switches app 
5. the user pastes the old username 
6. the user switches app back to Lockbox.
6a. the credentials on the screen are updated
7. the user copies the new password 
8. the user switches app
9. the user pastes the new password.

This happens in both schemes.

jhugman: I've added an `assertOnUiThread()` method in a new file `DebugSupport`.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
jhugman: 👍

PR #131, opened 2018-10-05
idea sharing

### LINKED ISSUE #103: Create shared idea file strategy
- [ ] license files
- [ ] ktlint rules
- [ ] optimize imports (don't wildcard)

sashei:
resolves #103


sashei:
@devinreams @jhugman @linuxwolf I haven't set up much in the way of local linting / inspections, so I'm welcoming other commits on this PR with local configuration that we believe should be shared.

thanks!!

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,29 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <Objective-C-extensions>
```
jhugman: Wait, wut?

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,24 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="AndroidLintAppLinksAutoVerifyError" enabled="true" level="ERROR" enabled_by_default="true" />
+    <inspection_tool class="AndroidLintAppLinksAutoVerifyWarning" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="AndroidLintBackButton" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="AndroidLintConvertToWebp" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="AndroidLintEasterEgg" enabled="true" level="WARNING" enabled_by_default="true" />
```
jhugman: …

.idea/copyright/Mozilla.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="myName" value="Mozilla" />
+    <option name="notice" value="This Source Code Form is subject to the terms of the Mozilla Public&#10;License, v. 2.0. If a copy of the MPL was not distributed with this&#10;file, You can obtain one at http://mozilla.org/MPL/2.0/." />
+  </copyright>
+</component>
```
jhugman: This is my most important setting.

linuxwolf:
Thanks for starting this @sashei.  I don't have any settings in Android Studio that aren't reflected here.

devinreams:
If I run 

```sh
ktlint --apply-to-idea-project --android 
```

from [the instructions](https://github.com/shyiko/ktlint#-with-intellij-idea) then I get a few more changes to the code styles config. It also sets these rules to be per-profile.

What do you think about including those here as part of this? I don't want to consider committing it here or in a separate branch until someone else looks at it and possibly tries it out during day-to-day-development.

.idea/codeStyles/Project.xml
sashei: I wonder if it pulled this from my AppCode config... I don't feel like we need to include this here....

sashei:
@devinreams I'll run that script locally and take a look. My feelings is that we should include the settings you're talking about and get this merged to `master`. We can then easily assess what's useful or not and make follow-up PRs to tweak the config as time goes on.

.idea/inspectionProfiles/Project_Default.xml
sashei: 🥚

sashei:
@devinreams I think those changes are fine; my only concern is that folks starting up with the project won't necessarily have `ktlint` installed locally. Can we add that to the bootstrap script if we're going to require its use in our config?

sashei:
follow-up: we do not have a bootstrap script :p

.idea/codeStyles/Project.xml
linuxwolf: I looked at my existing AS config, and it has this element.  It might be done (over) re-use on the name

sashei: ok, leaving it.

I am unable to delete it without AS putting it back so we have no choice.

bootstrap.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint &&
```
linuxwolf: Bootstrapping `ktlint` like this is problematic for a couple of reasons:

* It assumes a specific OS layout
* It potentially overrides other installs of `ktlint`

While it's only bootstrapping `ktlint` for now, it's safe enough to ignore this if one knows it's already installed.  If we have other items to bootstrap, this will need to be reworked.

sashei: Ok. I'm just blinding copying the `brew`-independent install from the `ktlint` documentation; I think it would be valid to update our `install.md` document with required libraries, rather than have this as part of `bootstrap.sh`.

.idea/inspectionProfiles/Project_Default.xml
devinreams: it's not an easter egg if you tell me about it... yeesh 🐰

PR #133, opened 2018-10-08
Add navigation graph to aid with persisting backstack.

### LINKED ISSUE #102: Revisit route and backstack persistence

jhugman:
Fixes #102.

This PR introduces [Android's navigation component][1] into the project. It is a significant refactor of the `RoutePresenter` and mitigates or eradicates many of the lifecycle/configuration change problems we have been seeing.

Much navigation complexity disappears into the `nav_graph` XML file. Adding new screens and transitions between screens should be done here. 

For the cases of app drawer menus, and fixed titles of toolbars, this need only be done in these graph files. Indeed, much of the wiring we needed to do for the `fragment_backable` and friends, and the `appDrawer` menu goes away.

The `RoutePresenter` is still required: it's largest job is to find the edge of the graph we're currently on.

We are likely going to need to add to this graph, and perhaps split it out into multiple graphs.

This PR also bulks out the UI testing, to include checking the behaviour of the back button.

#### Developer/testing instructions: 

As well as a gradle resync, you'll need to switch on the navigation editor in Android Studio:

> If using a Beta, Release Candidate, or Stable build, you must enable the Navigation Editor. Click File > Settings (Android Studio > Preferences on Mac), select the Experimental category in the left pane, check Enable Navigation Editor, and then restart Android Studio.

[1]: https://developer.android.com/topic/libraries/architecture/navigation/


sashei:
✨ linter errors ✨

app/src/main/java/mozilla/lockbox/presenter/BackablePresenter.kt
```diff
@@ -1,22 +1,17 @@
...
 
-import io.reactivex.Observable
-import io.reactivex.rxkotlin.addTo
-import mozilla.lockbox.action.RouteAction
 import mozilla.lockbox.flux.Dispatcher
 import mozilla.lockbox.flux.Presenter
 
-interface BackableViewProtocol {
-    val backButtonTaps: Observable<Unit>
+interface BackableView {
```
sashei: 👍 ~~`Protocol`~~

app/src/main/java/mozilla/lockbox/presenter/BackablePresenter.kt
```diff
@@ -1,22 +1,17 @@
...
 class BackablePresenter(
-    val view: BackableViewProtocol,
+    val view: BackableView,
     val dispatcher: Dispatcher = Dispatcher.shared
 ) : Presenter() {
     override fun onViewReady() {
-        view.backButtonTaps
-                .subscribe { dispatcher.dispatch(RouteAction.Back) }
-                .addTo(compositeDisposable)
+        // NOOP
```
sashei: do we need to override this function?

app/src/main/java/mozilla/lockbox/view/FilterFragment.kt
```diff
@@ -37,7 +37,6 @@ class FilterFragment : BackableFragment(), FilterView {
     ): View? {
         presenter = FilterPresenter(this)
         val view = inflater.inflate(R.layout.fragment_filter, container, false)
-        setupBackable(view, R.drawable.ic_arrow_back)
```
sashei: yayyyy getting rid of this method!

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -6,75 +6,85 @@
...
+                        .toBundle()
+                replaceFragment(R.id.fragment_item_detail, bundle)
             }
-            is RouteAction.Back -> activity.supportFragmentManager.popBackStack()
+            is RouteAction.Back -> navController.popBackStack()
         }
     }
+
+    private fun findTransitionId(@IdRes from: Int, @IdRes to: Int): Int? {
+        when (from) {
```
sashei: is there a cleaner way to encode the `from` and `to` transitions than this nested `when` structure?

sashei: I'm thinking `when(from, to)` and then we can construct the pairs as we construct the actions.

it's mostly the nesting that bugs me here.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -6,75 +6,85 @@
...
-            val base = fm.getBackStackEntryAt(0)
-            fm.popBackStackImmediate(base.id, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+    private fun replaceFragment(@IdRes destinationId: Int, args: Bundle? = null) {
+        val from = navController.currentDestination?.id ?: return
+        if (from == destinationId && args == null) {
+            // No point in navigating if nothing has changed.
+            return
         }
+        val transition = findTransitionId(from, destinationId) ?: destinationId
+        navController.navigate(transition, args)
```
sashei: hmm, so just to confirm that I'm reading this correctly... we navigate to the destination even if we don't have a specified transition for it?

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: Oh yes, that's a good idea. This is the code I'm least in love with.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: Correct. If we can't find a transition, then we can still get there, just not with a funky transition. We should have the funky transition. 

Should we `IllegalStateException` if not, to force us to keep the graph in sync?

sashei: I do like the funky transitions...

Being able to start views that aren't specified as "next" ones for the current view seems like it might be useful for things like deep-linking, though. If we are unlocked and on the settings page, but someone launches the app via some kind of autofill dialog, we still want to show the entries list or a specified entry.

My follow-up question is, what happens to the graph when it is not N'SYNC ("in sync")?

app/src/main/java/mozilla/lockbox/presenter/BackablePresenter.kt
jhugman: I deleted the method, but I'm leaving this mostly for a tech debt exercise namely: "What is the point of `BackablePresenter`?".

app/src/main/java/mozilla/lockbox/view/FilterFragment.kt
jhugman: :)

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: Not wild about this, but this is more in keeping with what you were after.

```kt
when (Pair(from, to)) {
    Pair(R.id.fragment_welcome, R.id.fragment_fxa_login) -> return R.id.action_welcome_to_fxaLogin
    …
}
```

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: There's a whole bunch of support for deep-linking, and putting together a back stack based on a deep-link, it's pretty cool. 

Also: we're going to want to be using multiple graphs, and multiple activities, and be able to navigate around them. It gets suuper complicated quite quickly.

So: in answer to your question, if the `RoutePresenter` is not in sync with the graph, then we just navigate straight to the destination.

I've improved the error message for you though.

jhugman:
Followup bugs filed.

sashei:
@jhugman I forgot to mention this before you merged, but I would appreciate some really basic documentation about needing to enable the "Experimental" feature in Android Studio at the very least, and possibly how to look at the `nav_graph.xml` file and add new screens.

PR #136, opened 2018-10-09
add delay check for clear clipboard username/password if still present

### LINKED ISSUE #49: Timer for Copy Action (detail view)
## Acceptance criteria
* Contents copied to the clipboard should expire after 60 seconds

Depends on #13

mihainisipeanusv:
Fixes #49 

add timer for clear username or password added in clipboard only if still present, if any other clipboard entry was added after is not cleared

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
@@ -51,16 +51,19 @@ open class ClipboardStore(
     private fun addToClipboard(label: String, str: String) {
         val clip = ClipData.newPlainText(label, str)
         clipboardManager.primaryClip = clip
-        clearClipboard(clipDataClearDelay, str)
+        timedReplaceDirty("", str)
     }
 
-    private fun clearClipboard(delay: Long, clearStr: String) {
+    private fun timedReplaceDirty(dirty: String, clean: String, delay: Long = clipDataClearDelay){
         Handler().postDelayed({
-            val clipData = clipboardManager.primaryClip.getItemAt(0)
-            if (clipData.text == clearStr) {
-
-                clipboardManager.primaryClip = ClipData.newPlainText("", "")
-            }
+            replaceDirty(dirty, clean)
         }, delay)
+    }
+
+    fun replaceDirty(dirty: String, clean: String) {
+        val clipData = clipboardManager.primaryClip.getItemAt(0)
+        if (clipData.text == clean) {
+            clipboardManager.primaryClip = ClipData.newPlainText("", dirty)
+        }
     }
 }
--- a/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
@@ -9,7 +9,6 @@ package mozilla.lockbox.store
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.os.Handler
-import android.util.Log
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.addTo
 import mozilla.lockbox.action.ClipboardAction
@@ -24,7 +23,7 @@ open class ClipboardStore(
         val shared = ClipboardStore()
     }
 
-    private val clipDataClearDelay = 60000L
+    private val defaultClipboardTimeout = 60000L
     private lateinit var clipboardManager: ClipboardManager
 
     init {
@@ -48,22 +47,22 @@ open class ClipboardStore(
         clipboardManager = manager
     }
 
-    private fun addToClipboard(label: String, str: String) {
-        val clip = ClipData.newPlainText(label, str)
+    private fun addToClipboard(label: String, string: String) {
+        val clip = ClipData.newPlainText(label, string)
         clipboardManager.primaryClip = clip
-        timedReplaceDirty(str)
+        timedReplaceDirty(string)
     }
 
-    private fun timedReplaceDirty(dirty: String, delay: Long = clipDataClearDelay){
+    private fun timedReplaceDirty(dirty: String, clean: String = "", delay: Long = defaultClipboardTimeout) {
         Handler().postDelayed({
-            replaceDirty(dirty)
+            replaceDirty(dirty, clean)
         }, delay)
     }
 
-    fun replaceDirty(dirty: String) {
+    fun replaceDirty(dirty: String, clean: String = "") {
         val clipData = clipboardManager.primaryClip.getItemAt(0)
         if (clipData.text == dirty) {
-            clipboardManager.primaryClip = ClipData.newPlainText("", "")
+            clipboardManager.primaryClip = ClipData.newPlainText("", clean)
         }
     }
 }
--- a/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
@@ -51,19 +51,19 @@ open class ClipboardStore(
     private fun addToClipboard(label: String, str: String) {
         val clip = ClipData.newPlainText(label, str)
         clipboardManager.primaryClip = clip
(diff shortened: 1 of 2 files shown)


app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -48,5 +51,16 @@ open class ClipboardStore(
     private fun addToClipboard(label: String, str: String) {
         val clip = ClipData.newPlainText(label, str)
         clipboardManager.primaryClip = clip
+        clearClipboard(clipDataClearDelay, str)
+    }
+
+    private fun clearClipboard(delay: Long, clearStr: String) {
```
jhugman: Rename these variables to fit the intent of the method.

Note that the method is approximately: clear the clipboard after a delay if and only if the clipboard contains this string.

Suggest: 

```kt
fun timedReplaceDirty(dirty: String, clean: String, delay: Int = clipDataClearDelay) {}

fun replaceDirty(dirty: String, clean: String) {}
```

Note:

 * since the class is called `ClipboardStore`, you don't have to have `Clipboard` in every identifier name.
 * you can test the second method pretty directly.
 * since you're calling the clear methods at the same time as the copy, you can restore the clipboard back to its original state after the credentials have been copied. This will require careful bookkeeping, so this may be worth discussing with product before doing this.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -48,5 +51,16 @@ open class ClipboardStore(
...
+        Handler().postDelayed({
+            val clipData = clipboardManager.primaryClip.getItemAt(0)
+            if (clipData.text == clearStr) {
+
```
jhugman: Nit: whitespace.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -48,5 +51,16 @@ open class ClipboardStore(
     private fun addToClipboard(label: String, str: String) {
         val clip = ClipData.newPlainText(label, str)
         clipboardManager.primaryClip = clip
+        clearClipboard(clipDataClearDelay, str)
+    }
+
+    private fun clearClipboard(delay: Long, clearStr: String) {
+        Handler().postDelayed({
```
jhugman: How does a `Handler` behave when the app is backgrounded or killed? The answer to this may have privacy/security implications that @linuxwolf may be interested in.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -48,5 +51,16 @@ open class ClipboardStore(
...
+            val clipData = clipboardManager.primaryClip.getItemAt(0)
+            if (clipData.text == clearStr) {
+
+                clipboardManager.primaryClip = ClipData.newPlainText("", "")
+            }
+        }, delay)
```
jhugman: Tests?

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
mihainisipeanusv: code updated with changes

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
mihainisipeanusv: if app is in background handler is working, if app is killed everything is killed including that handler

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
mihainisipeanusv: added test

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
mihainisipeanusv: removed

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
linuxwolf: There are limits to how much we can/should keep ourselves alive.  Absent a more system-managed mechanism (such as iOS has), this seems the best compromise between respecting user information and respecting the user's device resources (battery/CPU/memory).

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -51,16 +51,19 @@ open class ClipboardStore(
...
-            }
+            replaceDirty(dirty, clean)
         }, delay)
     }
+
+    fun replaceDirty(dirty: String, clean: String) {
+        val clipData = clipboardManager.primaryClip.getItemAt(0)
+        if (clipData.text == clean) {
+            clipboardManager.primaryClip = ClipData.newPlainText("", dirty)
+        }
```
jhugman:
```kt
if (clipData.text == dirty) {
    clipboardManager.primaryClip = ClipData.newPlainText("", clean)
}
```

You replace the dirty with the clean.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
jhugman: Exactly the responses I was hoping for. Thanks

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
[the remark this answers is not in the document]
mihainisipeanusv: The idea was to check if the current clipboard entry is the same as the one you try to remove it, and if yes then replace with a dirty one in this case "". This prevent that if in meantime the user added something else (related or not with this application) in the clipboard it will not be deleted

so i test that current clipboard entry is the same as the one added (username/password) and if yes only then update clipboard with dirty

app/src/test/java/mozilla/lockbox/store/ClipboardStoreTest.kt
```diff
@@ -58,4 +58,19 @@ class ClipboardStoreTest : DisposingTest() {
         val clipboardManager: ClipboardManager = RuntimeEnvironment.application.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
         assertTrue(clipboardManager.primaryClip.getItemAt(0).text.equals(testString))
     }
+
+    @Test
+    fun testReplaceDirty() {
+        val testString = "my_test_password"
+        val dirtyString = ""
```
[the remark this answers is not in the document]
mihainisipeanusv: you can say that dirty is the one that will be erased however you can say that dirty is the one that replace the clean one with what you want in order to prevent that someone will still that password

Depend on what angle you want to see the problem.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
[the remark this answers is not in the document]
linuxwolf: A couple of comments:

1) This phrasing is very confusing.  Most people expect to be checking for `dirty` and changing it to `clean`.  This function does the exact opposite, which is going to lead to bugs in our future.
2) I don't understand why there needs to be two parameters to this function; shouldn't the reset clip value always be ""?

app/src/test/java/mozilla/lockbox/store/ClipboardStoreTest.kt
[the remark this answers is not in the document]
linuxwolf: See comment on `ClipboardStore`.  This phrasing if very confusing, which is going to lead to bugs in the future.

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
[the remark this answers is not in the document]
jhugman: 1. Agreed. 
2. The clean string was a nod to a possible future where the clipboard's initial value is _restored_ rather than wiped. The `clean` string now defaults to an empty string.

Issue #135, opened 2018-10-08
Navigation drawer appearance


eliserichards:
Match design/style on Invision mockup

Invision: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/screens/309979874
Zeplin: https://app.zeplin.io/project/5b6895bfe4af825140aa8dbc/screen/5b6915ca008a822d0c4bc321

Depends on #11 (FxA user, name, email avatar)

Issue #37, opened 2018-07-12
Hook up FAQ links across the app


sandysage:
There are several spots throughout the app that link off to the internal FAQ for support.

Let's use this description to identify 1) the locations in the app that need to link to the FAQ and 2) the exact URL to link to
- [x] **Welcome**
Design: https://zpl.io/VDzrg4v
Link Text: Learn more
Link URL: https://lockbox.firefox.com/faq.html#how-do-i-get-my-saved-logins-into-firefox-lockbox

- [ ] **You're all Set** (onboarding confirmation)
Design: https://zpl.io/V0yxoJQ
Link Text: 256-bit encryption
Link URL: https://lockbox.firefox.com/faq.html#what-security-technologies-does-firefox-lockbox-use 

- [x] **No Entries Found**
Design: https://zpl.io/bAqXRlB
Link Text: Learn more
Link URL: https://lockbox.firefox.com/faq.html#how-do-i-enable-sync-on-firefox

- [x] **No Matching Entries** (when searching)
Design: https://zpl.io/2j4XN1q
Link Text: Learn more
Link URL: https://lockbox.firefox.com/faq.html#how-do-i-create-new-entries

- [x] **Entry Detail**
Design: https://zpl.io/2EAZQyl
Link Text: Learn how to edit this entry
Link URL: https://lockbox.firefox.com/faq.html#how-do-i-edit-existing-entries

- [x] **Send Usage Data**
Design: https://zpl.io/awYA3LK
Link Test: Learn more
Link URL: https://lockbox.firefox.com/privacy.html

blocked by #73

changecourse:
@linuxwolf added!

sandysage:
@changecourse @nickbrandt I'm removing the "needs design" label and marking as P2 to address prior to public launch. Let me know if I made this change in error.

sashei:
@nickbrandt @changecourse just confirming, the title for _all_ of these pages is FAQ, regardless of what portion of the FAQ they link to?

thanks

nickbrandt:
@sashei yep that’s correct!

Issue #120, opened 2018-10-02
Record settings telemetry events


linuxwolf:
Depends on #37 
Depends on #6 

Record telemetry events for:

* ~The Settings view is displayed~ (done via #259)
* ~The FAQ is opened~ (done via #259)
* ~The Provide Feedback is opened~ (done via #259)
* The autolock timer option list is expanded 
* A setting is changed

Blocks #14

devinreams:
@linuxwolf @irrationalagent sorry but I didn't quite follow the resulting back and forth about all route actions being telemetry actions. does #259  mean these setting events are completed? I don't think so but I also don't think this is "in progress" if not...

https://github.com/mozilla-lockbox/lockbox-android/blob/fab63bbc495dd1116d5c995f41aa3b9ee4760151/app/src/main/java/mozilla/lockbox/action/RouteAction.kt#L21-L23

irrationalagent:
yea backlog is right, those events were only for the setting views not the events associated with setting changes, not started on the proper settings events yet.

PR #138, opened 2018-10-09
Setting view

### LINKED ISSUE #6: Settings view
When I want to customize my in app experience, I need to easily find those settings clear instructions for how to curate my in-app settings as well as delete my account.

## Acceptance criteria
- [x] Should be able to close out of the settings view to get back to the entry list
- [x] Should be easily accessible from Navigation Drawer
- [x] Should include the setting option list of
   - [x] Turn On/Off Fingerprint
   - [x] Change Auto Lock timer
   - [x] Turn On/Off AutoFill (this will actually just display whether the setting is turned on/off in settings. tapping this toggle would route the user to the setting on their device)
   - [x] Turn On/Off Sending Usage Data
   - [x] Display the App Version

### Visual Design

**InVision**
https://mozilla.invisionapp.com/share/3SN35UHWJDX#/309979885_Settings

**Zeplin**
https://zpl.io/awYA3LK

Remaining engineering todos:
- [x] toggle cell with link
- [x] app version
- [x] force cast in  `SettingFragment`
- [x] SettingPresenter test
- [x] settingListAdapter test
- [x] sectionlistadapter test
- [x] check bitrise

sashei:
resolves #6 

testing notes:

- we have created a follow-up story #135 to address the visual work on the navigation drawer
- all settings behavior will be handled in follow-up stories
- we are not hooking up the Learn More button yet, there is a separate story for all Learn More buttons: #37


app/src/main/java/mozilla/lockbox/presenter/SettingPresenter.kt
```diff
@@ -6,11 +6,33 @@
...
+}
 
-class SettingPresenter(view: SettingViewProtocol) : Presenter() {
+class SettingPresenter(val view: SettingView) : Presenter() {
     override fun onViewReady() {
+        val settings = listOf(
+            ToggleSettingConfiguration("Unlock with fingerprint", toggle = false),
+            TextSettingConfiguration("Auto lock", detailText = "5 minutes"),
+            ToggleSettingConfiguration("Autofill", "Let Firefox Lockbox fill in logins for you", false),
+            ToggleSettingConfiguration("Send usage data", "Mozilla strives to only collect what we need to provide and improve Firefox for everyone. ", true)
```
linuxwolf: (Yes, I realize this is still WIP)

When putting in the TelemetryStore basics, included resource strings/bools in place.  Is it reasonable to hook that in here now?

linuxwolf: #120 is to record Telemetry events around settings, not hooking in the "send usage data" setting

sashei: Oh, you're referring to story #8

linuxwolf: Yes, that's the one!  Might be nice to sneak it in now anyway, but please carry on! (-:

app/src/main/java/mozilla/lockbox/presenter/SettingPresenter.kt
```diff
@@ -6,11 +6,54 @@
...
+                toggle = false
+            ),
+            ToggleSettingConfiguration(
+                title = "Send usage data",
+                subtitle = "Mozilla strives to only collect what we need to provide and improve Firefox for everyone. ",
+                buttonTitle = "Learn more",
+                toggle = true
+            ),
+            AppVersionSettingConfiguration(
+                text = "App Version: $versionNumber"
```
jhugman: This bothers me: there isn't any use of `R.string` so far, and if we're dealing with Strings in this `Presenter` then we need to have a `Context.getResources()`, which goes against the separation of view from presenter.

Suggest you switch the configuration objects to use `@StringRes Int`, with a `res/values/settings.xml` strings file.

app/src/main/res/layout/list_cell_setting_text.xml
```diff
@@ -0,0 +1,40 @@
...
+        android:textColor="@color/opaque_black"
+        android:letterSpacing="0.03"
+        android:lineSpacingExtra="12sp"
+        android:layout_marginStart="16dp"
+        android:layout_marginTop="12dp"
+        android:layout_marginBottom="12dp"
+        tools:text="Auto lock"/>
+
+    <TextView
+            android:id="@+id/settingSelection"
```
jhugman: Nit: indent should match L10-L22.

app/src/main/res/layout/fragment_setting.xml
```diff
@@ -4,27 +4,19 @@
   ~ License, v. 2.0. If a copy of the MPL was not distributed with this
   ~ file, You can obtain one at http://mozilla.org/MPL/2.0/.
   -->
-<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    tools:context=".view.SettingFragment">
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
jhugman: There's a lot of mixing between layouts here. Sometimes `RelativeLayout`, sometimes `ConstraintLayout`, sometimes `LinearLayout`.

Suggest we use `FrameLayout`s for layouts with one view, and `ConstraintLayout` everywhere else.

app/src/main/java/mozilla/lockbox/adapter/SettingCellConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+class ToggleSettingConfiguration(
+    override val title: String,
+    override val subtitle: String? = null,
+    val buttonTitle: String? = null,
+    val toggle: Boolean
+) : SettingCellConfiguration(title)
+
+class AppVersionSettingConfiguration(
+    val text: String
+) : SettingCellConfiguration(text)
```
jhugman: A little confused why we didn't use [PreferenceFragmentCompat][1] here.

[1]: https://developer.android.com/reference/android/support/v7/preference/PreferenceFragmentCompat

app/src/test/java/mozilla/lockbox/adapter/SectionedAdapterTest.kt
```diff
@@ -0,0 +1,212 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(packageName = "mozilla.lockbox")
+class SectionedAdapterTest {
+
```
jhugman: Nit: whitespace.

app/src/main/java/mozilla/lockbox/view/SettingFragment.kt
```diff
@@ -23,6 +38,28 @@ class SettingFragment : BackableFragment(), SettingViewProtocol {
         presenter = SettingPresenter(this)
 
         // Inflate the layout for this fragment
-        return inflater.inflate(R.layout.fragment_setting, container, false)
+        val view = inflater.inflate(R.layout.fragment_setting, container, false)
+
```
jhugman: Nit: Set `android:overScrollMode="ifContentScrolls"` in the `RecyclerView` xml please.

app/src/main/java/mozilla/lockbox/adapter/SectionedAdapter.kt
```diff
@@ -0,0 +1,115 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+// Large portions of this class are borrowed from:
+// https://gist.github.com/gabrielemariotti/4c189fb1124df4556058
```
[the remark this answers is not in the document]
sashei: @linuxwolf do you have any thoughts here?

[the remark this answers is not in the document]
devinreams: I see someone asked the author about licensing once already...

Would you all mind trying to reach out via another GitHub comment and perhaps email (if available/findable)? It definitely gets a bit sticky here but the "spirit" of the sharing of the code is definitely for others to re-use it.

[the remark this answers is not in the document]
linuxwolf: The lack of a license is concerning; we might need to include someone more knowledgeable on licensing.

Generally, though, code (heavily or lightly) inspired by another codebase can be relicensed if the original source license is compatible (for MPL, that's just about everything).  Since the license is not stated in the gist or its included files, was there more information from sources that led to this gist (and most importantly, are they stable enough to use as a reference)?

eliserichards:
I’ve been trying to get @jhugman’s suggestion about using `PreferenceFragmentCompat` for this view (instead of a hardcoded approach, like initial implementation).

After taking a deeper look, I don’t think that `PreferenceFragmentCompat` is going to work very well for our needs. It is simple to implement and takes care of a lot of the overhead for us, but it is *very* unstylable. I tried to find a way around this styling today, but this was the best I could do (this version inherits all styling from the base `AppTheme` style, and can only have one other style applied to it):
image

That being said, I think that it would be better to continue with the implementation that @sasha and I created, which is simpler and has styling which more closely matches what we want:
image

eliserichards:
I addressed @jhugman 's comments above.

@nickbrandt -- some of your suggested changes (e.g. white toggle when "off", 30px padding before the second section) are too involved to be included in this scope. They aren't super easy changes to make. I would suggest a followup story to clean up some of those small UI details.

app/src/main/res/layout/fragment_setting.xml
eliserichards: With the linear layout, the view is nested inside so that the "settings" header and back button are viewable. If this changes to a frame layout or constraint layout, the header and back button no longer show up. 
(the rest of the .xml files are LinearLayouts)

app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+        when (viewType) {
+            SETTING_TEXT_TYPE -> {
+                val view = inflater.inflate(R.layout.list_cell_setting_text, parent, false)
+                return TextSettingViewHolder(view)
+            }
+            SETTING_TOGGLE_TYPE -> {
+                val view = inflater.inflate(R.layout.list_cell_setting_toggle, parent, false)
+                return ToggleSettingViewHolder(view)
+            }
+            else -> {
```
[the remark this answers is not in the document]
eliserichards: not sure what the "last" case outside of the when should be. Throw an error/exception if it's not one of our three choices?

eliserichards:
@jhugman ready to review!

app/src/main/res/values/styles.xml
```diff
@@ -5,38 +5,46 @@
...
 
 <resources>
-
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <!-- Customize your theme here. -->
         <item name="colorPrimary">@color/colorPrimary</item>
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
-        <item name="android:textColor">#ffffff</item>
+        <item name="android:textColor">@color/text_white</item>
```
jhugman: YES!

app/src/main/java/mozilla/lockbox/adapter/SectionedAdapter.kt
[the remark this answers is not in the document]
devinreams: >Since the license is not stated in the gist or its included files, was there more information from sources that led to this gist

I investigated a bit more myself and the author has not explicitly explained the license situation. I've added this to our legal review checklist: #161

eliserichards:
Created new issue #163 (https://github.com/mozilla-lockbox/lockbox-android/issues/163) to address James' comments.

PR #141, opened 2018-10-10
Entry detail design

### LINKED ISSUE #123: Remaining entry detail tasks
Follow up to #112 

This covers some of the remaining work to be done on the entry detail page, outside of #12 (reveal) & #13 (copy)

**Elements**
- [x] "Learn how to edit this entry" Link (placeholder until #37 )

**Styling**
- [x] adjust app bar to have blue background
- [x] update "back" icon
- [x] put fields on a "card"
- [x] adjust margins & spacing
- [x] typography styles
- [x] input/border styles

### Visual Design

**InVision**

**Zeplin**
https://zpl.io/2EAZQyl

mihainisipeanusv:
Fixes #123 

Adding remaining styles for entry detail fragment


app/build.gradle
```diff
@@ -57,6 +57,7 @@ dependencies {
     implementation 'com.android.support.constraint:constraint-layout:1.1.3'
     implementation 'com.android.support:design:27.1.1'
     implementation 'com.android.support:recyclerview-v7:27.1.1'
+    implementation 'com.android.support:cardview-v7:27.1.1'
```
jhugman: At some point, we should add variable for 
```groovy
support_version = '27.1.1'
```

jhugman:
Adding @changecourse and @nickbrandt for UI review.

app/build.gradle
mihainisipeanusv: I was thinking at that but I did not want to take this in my own hands without asking

linuxwolf: I agree with the sentiment @jhugman.  It might be premature to do that, as these support packages are now effectively being replaced with [AndroidX](https://developer.android.com/jetpack/androidx/).  I think it's ok to hold off _for now_, since these values **MUST** match the `compileSDK` version or `gradlew check` complains loudly.

I have a branch I've been looking into some of these migrations, and will submit an issue to really track it.

linuxwolf:
Here's a screencap 💟 :

image

sashei:
@nickbrandt I didn't see the badness with the EditText underline color (that's the line underneath the values for Username, Password, etc) when building, so possibly that was from an older build or has something to do with my device in particular. let me know how you go with this version!

Issue #140, opened 2018-10-09
Explore Leak Detection


linuxwolf:
Explore options for detecting and reporting memory and/or resource leaks.

Ideally the solution becomes part of CI builds and tests.

Possible options:

* [LeakCanary](https://github.com/square/leakcanary)

linuxwolf:
Something that came up during an IRL review.

@mozilla-lockbox/mobile-engineering drop any suggestions!

sashei:
+1 for LeakCanary, was really useful in a previous app I've worked on, and plays well with Rx.

There is a static analyzer that Facebook makes (leakcanary is made by Square) called [Infer](https://fbinfer.com/docs/getting-started.html) that I have no experience with, but could be useful as a CI tool (IIRC LeakCanary only really works when you're looking at the emulator or device while using the app).

devinreams:
With #142 merged shall we consider this "complete" or leave it open (back in the backlog? inbox?) to remind us to explore/review further?

linuxwolf:
[ cc all of @mozilla-lockbox/mobile-engineering ]

I lean toward leaving it open for a sprint, and see how it works out.

sashei:
^ jinx on the `cc`

I feel that future issues with LeakCanary should be addressed separately from this story. There isn't much definition in the ticket, but I left it open because we may want to include leak detection / monitoring at a CI level. LeakCanary only works on devices / emulators when a user is looking at them.

linuxwolf:
> I feel that future issues with LeakCanary should be addressed separately from this story. There isn't much definition in the ticket, but I left it open because we may want to include leak detection / monitoring at a CI level. LeakCanary only works on devices / emulators when a user is looking at them.

I agree about future issues, but I think it's also worth keeping open for a bit to see if it's actually useful.  I would like CI gatekeepings, assuming they're useful.

devinreams:
In advance of "finishing" this sprint tomorrow, consider this a prompt for any initial thoughts or observations during this sprint (if any).

devinreams:
Follow up at #172 to add to CI

PR #142, opened 2018-10-11
add leak canary to app

sashei:
connected to #140

### DIFF
--- a/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
+++ b/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
@@ -9,6 +9,7 @@ package mozilla.lockbox
 import android.app.Application
 import android.content.ClipboardManager
 import android.content.Context
+import com.squareup.leakcanary.LeakCanary
 import io.sentry.Sentry
 import io.sentry.android.AndroidSentryClientFactory
 import mozilla.components.support.base.log.Log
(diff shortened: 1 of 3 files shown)


app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -21,6 +22,13 @@ val log: Logger = Logger("Lockbox")
 class LockboxApplication : Application() {
     override fun onCreate() {
         super.onCreate()
+        if (LeakCanary.isInAnalyzerProcess(this)) {
+            // This process is dedicated to LeakCanary for heap analysis.
+            // You should not init your app in this process.
+            return
+        }
+        LeakCanary.install(this)
+
```
jhugman: Should this whole block be wrapped in 

```kt
if (BuildConfig.DEBUG) {}
```

sashei: No, because in `build.gradle` we do the `noop` version of LeakCanary for the `releaseImplementation` configuration.

sashei:
@jhugman we can't ship this just yet because we need to disable this for unit tests.

docs/contributing.md
```diff
@@ -12,6 +12,10 @@ This repository is governed by Mozilla's code of conduct and etiquette guideline
 
 Please refer to installation and build instructions in the [documentation](install.md).
 
+## LeakCanary
+
+We use [LeakCanary](https://github.com/square/leakcanary/wiki/FAQ) to monitor memory leaks in Debug builds of the Lockbox Android application. Please read the documentation and pay attention to warnings while debugging the application!
```
linuxwolf: 👀 👍

PR #145, opened 2018-10-11
Application Lifecycle Telemetry

### LINKED ISSUE #119: Record app lifecycle telemetry events
Record telemetry events for:

* The app comes to the foreground
* The app goes into the background

linuxwolf:
Resolves #119


app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt
```diff
@@ -9,11 +9,13 @@ package mozilla.lockbox.action
 import mozilla.lockbox.flux.Action
 import org.mozilla.telemetry.event.TelemetryEvent
 
-abstract class TelemetryAction : Action {
-    abstract val eventMethod: TelemetryEventMethod
-    abstract val eventObject: TelemetryEventObject
-    open val value: String? get() = null
-    open val extras: Map<String, Any>? = null
+interface TelemetryAction : Action {
```
linuxwolf: @sashei you were right

linuxwolf:
[ cc @irrationalagent ]

Here's the first realworld telemetry eventing.  Let's see if this works for a starting point.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -91,4 +93,14 @@ class RoutePresenter(
...
+
+    override fun onPause() {
+        super.onPause()
+        Dispatcher.shared.dispatch(LifecycleAction.Background)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        Dispatcher.shared.dispatch(LifecycleAction.Foreground)
+    }
```
jhugman: This might work for now, however, it's extremely extremely brittle:

 * in the event of a configuration change (for example but not limited to: orientation change, car mode, night mode etc), `onPause` will be fired immediately followed by `onResume`.
 * when we add more activities (we will do shortly), the same will happen.

Suggest: 

 * we have an `ApplicationPresenter`, which fires off these events.
 * we use a [`ProcessLifecycleOwner`][1] and `LifecycleObserver` which does the OS specific detection, and hands off to the `ApplicationPresenter`. The [`ProcessLifecycleOwner`][1] is designed for this exact use case of detecting the app going into background and coming back in to foreground.

[1]: https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner

app/src/main/java/mozilla/lockbox/store/LifecycleStore.kt
```diff
@@ -0,0 +1,35 @@
...
+    val dispatcher: Dispatcher = Dispatcher.shared
+) {
+    companion object {
+        val shared = LifecycleStore()
+    }
+
+    internal val compositeDisposable = CompositeDisposable()
+    private val _filter: PublishSubject<LifecycleAction> = PublishSubject.create()
+
+    val lifecycelFilter: Observable<LifecycleAction> = _filter
```
jhugman: Nit: naming `lifecycleFilter`.

app/src/main/java/mozilla/lockbox/store/LifecycleStore.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    init {
+        dispatcher.register
+            .filterByType(LifecycleAction::class.java)
+            .subscribe { _filter.onNext(it) }
+            .addTo(compositeDisposable)
+    }
```
jhugman: `_filter` is redundant here.

Suggest `init` method:

```kt
    val lifecycelFilter: Observable<LifecycleAction> = PublishSubject.create()

    init {
        val subject = lifecycelFilter as PublishSubject
        dispatcher.register
            .filterByType(LifecycleAction::class.java)
            .subscribe(subject::onNext)
            .addTo(compositeDisposable)
    }
```

app/src/test/java/mozilla/lockbox/store/LifecycleStoreTest.kt
```diff
@@ -0,0 +1,35 @@
...
+package mozilla.lockbox.store
+
+import mozilla.lockbox.DisposingTest
+import mozilla.lockbox.action.LifecycleAction
+import mozilla.lockbox.flux.Dispatcher
+import org.junit.Test
+
+class LifecycleStoreTest : DisposingTest() {
+    @Test
+    fun testLifecycelFilter() {
```
jhugman: Nit: naming.

app/src/main/java/mozilla/lockbox/action/LifecycleAction.kt
```diff
@@ -0,0 +1,18 @@
...
+package mozilla.lockbox.action
+
+enum class LifecycleAction(
+    override val eventMethod: TelemetryEventMethod,
+    override val eventObject: TelemetryEventObject
+) : TelemetryAction {
+    Foreground(TelemetryEventMethod.foreground, TelemetryEventObject.app),
+    Background(TelemetryEventMethod.background, TelemetryEventObject.app),
+    Startup(TelemetryEventMethod.startup, TelemetryEventObject.app),
+    // TODO: Add a TelemetryEventMethod for upgrading
```
jhugman: 🙁

app/src/main/java/mozilla/lockbox/store/LifecycleStore.kt
linuxwolf: This is what we called it in iOS.

app/src/main/java/mozilla/lockbox/action/LifecycleAction.kt
linuxwolf: something to talk to @irrationalagent about for both iOS and Android.

app/src/main/java/mozilla/lockbox/presenter/ApplicationPresenter.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.arch.lifecycle.DefaultLifecycleObserver
+import android.arch.lifecycle.LifecycleOwner
+import mozilla.lockbox.action.LifecycleAction
+import mozilla.lockbox.flux.Dispatcher
+
+class ApplicationPresenter(
+    val dispatcher: Dispatcher = Dispatcher.shared
+) : DefaultLifecycleObserver {
+    override fun onCreate(owner: LifecycleOwner) {
+        Dispatcher.shared.dispatch(LifecycleAction.Startup)
```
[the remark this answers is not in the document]
linuxwolf: This was a dropping ... but I could do it that way

app/src/test/java/mozilla/lockbox/presenter/ApplicationPresenterTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    override fun removeObserver(observer: LifecycleObserver) {
+    }
+}
+
+class ApplicationPresenterTest : DisposingTest() {
+    private val dispatcherObserver = createTestObserver<Action>()
+
+    @Before
+    fun setUp() {
+        Dispatcher.shared.register.subscribe(dispatcherObserver)
```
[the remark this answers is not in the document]
linuxwolf: I was following a pattern mostly in place from other -`PresenterTest`s.  That said, I was working in more in this direction at first, so I can switch to it.

I was creating the `subject` per test thinking to avoid side-effects (which don't exist today).

Issue #68, opened 2018-08-28
Implement Biometric (fingerprint) unlock fragment/activity


linuxwolf:
Implement the Activity or FragmentActivity for [Unlock Firefox Lockbox](https://mozilla.invisionapp.com/share/3SN35UHWJDX#/screens/310606417_Lock_Screen_-_Fingerprint) design.

* Uses FingerprintManager API to engage hardware when initialized
* Cancel hardware when user clicks "Cancel"
* Dismiss (Fragment)Activity when hardware successfully unlocks via fingerprint
* Fallback to device PIN when the user requests it

Related #28 
Required by #31
Required by #32
Required by #69
Required by #70
Required by #118
Required by #144

ioana-farcas:
@nickbrandt , @changecourse 
Which should be the behaviour of the app when the fingerprint authentication fails?
If the fingerprint authentication fails for 5 times in a row, the fingerprint sensor will be unresponsive for 30 seconds, and the API sets a message for the error: "Too many attempts. Try again later" (the error message can be changed in the dialog).

Also in other apps (for unlocking the screen is the same), when the fingerprint authentication fails for 5 times in a row, the behaviour is to fallback to other type of authentication. If for this other type, in our case PIN, the authentication will fail again for 5 times in a row, the user will not be able to unlock app for 30 seconds. After those 30 seconds pass, the user can use fingerprint or pin to authenticate.

I added a screenshot with the dialog, please advise how do you think the designs and the UX should be/look like for this scenario and please let me know what do you thing about falling back to PIN, taking into account that the sensor will be unresponsive for 30 seconds when fingerprint auth fails.

![fingerprint_fail](https://user-images.githubusercontent.com/16756095/47139419-79948680-d2c4-11e8-8bcb-c1f0e7f642df.jpg)

changecourse:
My hope for how we handle this, based on a conversation with @linuxwolf would be as follows:

1. Present fingerprint auth for the first 5 attempts (solely)
2. Then direct user to Device PIN input automatically
3. After all attempts for both fingerprint auth and device pin have failed, present lockout messaging

How automated we are able to be (as referenced in no. 2 may need to still be up for discussion.

changecourse:
@nickbrandt there was a request in stand-up to confirm the following states (or approve the default styling by Android) 

1. Fingerprint unlock failed (too many attempts)
2. PIN unlock failed (too many attempts)
3. Lockout state

[...]

Issue #22, opened 2018-07-12
Signing in for the first time (Signing in and authenticating with FxA)


sandysage:
## Acceptance criteria
* Sign in should happen within the app view

## Tasks
- [x] [Request](https://bugzilla.mozilla.org/show_bug.cgi?id=1499481) OAuth client be provisioned ([Details](https://github.com/mozilla-lockbox/lockbox-android/issues/22#issuecomment-433131989))
- [x] Implement AccountStore using `service-firefox-accounts`
- [x] FxA login with `cancel` button

### Visual Design

**InVision**
FxA Auth (email first flow): https://mozilla.invisionapp.com/share/3SN35UHWJDX#/311586971_FxA_-_Email

**Zeplin**
FxA Auth (email first flow): https://zpl.io/blpW3DG

linuxwolf:
Request to provision OAuth client credentials tracked via [bugzilla 1499481](https://bugzilla.mozilla.org/show_bug.cgi?id=1499481)

sashei:
Some initial / storage work in #151

linuxwolf:
OAuth client provisioned:

```

{
  "id": "e7ce535d93522896",
  "name": "Firefox Lockbox for Android",
  "redirectUri": "https://lockbox.firefox.com/fxa/android-redirect.html",
  "canGrant": true,
  "trusted": true,
  "allowedScopes": "https://identity.mozilla.com/apps/lockbox
https://identity.mozilla.com/apps/oldsync",
  "publicClient": true
}
```

PR #151, opened 2018-10-16
securepreferences

### LINKED ISSUE #67: Integrate DataProtect library as a flux store
All sensitive information should be only be saved to the user's device encrypted.  Such data includes (but not necessarily limited to):

- ~~sqlcipher database encryption key~~ leaving for later until we understand Sync requirements better
- [x] FxA Account details (profile information, OAuth tokens)
- [ ] provide Cipher for Fingerprint Manager / Biometric Fragment

To accomplish this, this story is responsible for creating a `Store` with Observables for the relevant encrypted data and accompanying `Action`s for updating that data.

sashei:
resolves #67 

remaining work:
 - [x] updating the tests for `SecurePreferences` to properly mock the Keystore (conversation in a-c is here: mozilla-mobile/android-components#1072)
 - [x] seeing how this could be integrated with the work on the Fingerprint Manager (issue #68) using the Keystore helpers (waiting for @ioana-farcas to push work there)
 - [x] finishing up AccountStore class with the full FxA flow & proper unit tests (addressed in issue #22 but some work still required there. also difficult to test because we cannot run unit tests on the cross-compiled rust code. application-services issue here: mozilla/application-services#262)


app/src/main/java/mozilla/lockbox/store/SecureStore.kt
```diff
@@ -0,0 +1,97 @@
...
+    val oauthInfo: Observable<OAuthInfo?> = PublishSubject.create()
+    val profile: Observable<Profile?> = PublishSubject.create()
+
+    init {
+        this.dispatcher.register
+            .filterByType(SecureAction::class.java)
+            .subscribe {
+                when (it) {
+                    is SecureAction.OAuthInfo -> {
+                        // tbd: how to convert to & from String values for OAuthInfo...
```
linuxwolf: yeah yeah yeah, WIP (-:

This might provide some pointers for the `tbd`s: [samples/firefox-accounts](https://github.com/mozilla-mobile/android-components/blob/dbaf87110dbf5f7183ba067c086049f4b4ff5bb6/samples/firefox-accounts/src/main/java/org/mozilla/samples/fxa/MainActivity.kt#L137)

sashei: thanks matt :)

app/src/main/java/mozilla/lockbox/support/Optional.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.support
+
+data class Optional<T>(val value: T?)
+fun <T> T?.asOptional() = Optional(this)
```
jhugman: Ooh. Weird.

app/src/main/java/mozilla/lockbox/action/SecureAction.kt
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+package mozilla.lockbox.action
+
+import mozilla.components.service.fxa.FirefoxAccount
+import mozilla.lockbox.flux.Action
+
+sealed class SecureAction : Action {
+    data class FxAAccount(val account: FirefoxAccount) : SecureAction()
+}
```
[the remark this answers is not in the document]
sashei: I guess I was thinking you would "secure" your FxA account as an action, but yeah I can see the ambiguity.

app/src/main/java/mozilla/lockbox/support/Optional.kt
sashei: gotta make Rx play nicely with optionals...

sashei:
I realized in addressing these comments that there's been some scope drift in this PR. I've converted the `SecureStore` to an `AccountStore` to match iOS, and made a `SecurePreferences` helper class that actually addresses the ticket.

There's some remaining work, which includes:
- updating the tests to properly mock the `Keystore` and test the `SecurePreferences` class (conversation in `a-c` is here: https://github.com/mozilla-mobile/android-components/issues/1072)
- seeing how this could be integrated with the work on the Fingerprint Manager (issue #68) using the `Keystore` helpers (waiting for @ioana-farcas to push work there)
- finishing up `AccountStore` class with the full FxA flow & proper unit tests (addressed in issue #22 but some work still required there. also difficult to test because we cannot run unit tests on the cross-compiled rust code. `application-services` issue here: https://github.com/mozilla/application-services/issues/262)

sashei:
follow-up: we still need to make `Keystore` open because you cannot do the `mock-maker-inline` solution for instrumentation tests, and with the version of `Base64` from the examples, you cannot run this code _outside_ of an Android VM in unit tests D:

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    val oauthInfo: Observable<Optional<OAuthInfo>> = PublishSubject.create()
+    val profile: Observable<Optional<Profile>> = PublishSubject.create()
+
+    init {
+        // registration for further actions happens here
+    }
+
+    fun apply(sharedPreferences: SharedPreferences) {
+        securePreferences.apply(sharedPreferences)
```
jhugman: Pass in the `SecurePreferences` object, not the `SharedPreferences` object.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        loadFxA()
+    }
+
+    private fun loadFxA() {
```
jhugman: Nit: there's a bunch of things I don't like about this naming. 

First: prefer `FirefoxAccount` or `Account` to `FxA`.

Second: you're not returning anything (as you might expect to `load` an existing account); you're then resaving immediately after you've loaded, or updating the profile/oauth with `null`.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        val profileSubject = profile as PublishSubject
+        account.getProfile().whenComplete {
+            profileSubject.onNext(it.asOptional())
+        }
+
+        val oauthSubject = oauthInfo as PublishSubject
+        account.getOAuthToken(FXA_SCOPES).whenComplete {
+            oauthSubject.onNext(it.asOptional())
+        }
```
jhugman: Similarly, this method needs a rethink/rename. 

We'e not just `loading` or `persisting` in either of these methods. 

Perhaps they need to be split up into smaller methods.

app/src/main/java/mozilla/lockbox/support/Optional.kt
jhugman: Yeah, that sucks. Perhaps we shouldn't be using optionals in Rx?

Playing around, there don't seem to be any good answers to this.

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
```diff
@@ -0,0 +1,46 @@
...
+            val encrypted = Base64.decode(value, BASE_64_FLAGS)
+            val plain = keystore.decryptBytes(encrypted)
+
+            String(plain, StandardCharsets.UTF_8)
+        } else {
+            null
+        }
+    }
+
+    open fun putString(key: String, value: String) {
```
jhugman: Nit: try and make this mirror the `SharedPreferences` API as much as possible.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
sashei: That means `LockboxApplication` will be doing heavier lifting than before, but I suppose that's fine.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
sashei: We need to resave immediately after loading the `FirefoxAccount` from the JSON string because that's when they do the account refresh.

I'm mirroring what we have in iOS here, but I'm happy to refactor if it would be clearer.

sashei:
@jhugman I don't love the `apply` pattern either. What other ideas did we have to get context in stores where it's necessary?

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import android.content.SharedPreferences
+import android.util.Base64
+import mozilla.components.lib.dataprotect.Keystore
+import java.nio.charset.StandardCharsets
+
+private const val KEYSTORE_LABEL = "lockbox-keystore"
+private const val BASE_64_FLAGS = Base64.URL_SAFE or Base64.NO_PADDING
+
+open class SecurePreferences(
```
linuxwolf: At some point, `Keystore.generateKey()` must be called, or all the calls to encrypt/decrypt will error.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
linuxwolf: put me down for "not a fan of this change", unless `SecurePreferences` is maintained/initialized by `LockboxApplication` for other pieces.

linuxwolf:
> @jhugman I don't love the `apply` pattern either. What other ideas did we have to get context in stores where it's necessary?

Are we discussing this _again_? (-:

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
sashei: oh.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -0,0 +1,75 @@
...
+import mozilla.components.service.fxa.FirefoxAccount
+import mozilla.components.service.fxa.OAuthInfo
+import mozilla.components.service.fxa.Profile
+import mozilla.lockbox.support.Optional
+import mozilla.lockbox.support.asOptional
+import mozilla.lockbox.flux.Dispatcher
+import mozilla.lockbox.support.SecurePreferences
+
+private const val FIREFOX_ACCOUNT_KEY = "firefox-account"
+private val FXA_SCOPES = arrayOf("profile", "https://identity.mozilla.com/apps/lockbox")
```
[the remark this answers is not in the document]
sashei: why do we need oldsync for Android? I was confused about this when I saw your bugzilla conversation go by. I thought we only needed oldsync in iOS for the FxiOS Sync code and the application-services version will work with the new `sync`

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
linuxwolf: my slight bias is to do this lazily, at the point where encrypt/decrypt are first called:

* the provisioning is cheap, so the latency is minor
* it's arguably a better place in the app's lifecycle to deal with errors?
* If it doesn't exist when there's definitely a value to decrypt, we can differentiate the failure better (e.g., _something_ about the app left `SharedPreferences` around but not an encryption key, vs the decrypt failed because the keys are different)?

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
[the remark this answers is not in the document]
linuxwolf: That is not correct; the `oldsync` scope gets us the FxSync key.  Even the new `sync` implementation needs this same key, or nothing from FxSync will decrypt.

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
sashei: This is functionally the same, but I felt like the "editor" concept was pretty high overhead for a two method class.

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
sashei: Ok, I'd like to link up with you to talk through the error cases and what handling we should have here.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
sashei: :sparkles: _no one knows_ :sparkles:

sashei: I don't like any of the above tbh; I think we agreed to keep the `apply` pattern to get those other PRs landed, but I would rather come up with another solution because there will undoubtedly be many other `Context`-requiring `Store`s.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
sashei: Maybe we need the rename for `persist`. We should never save the `FirefoxAccount` without updating the `oauthInfo` and `profile` observables to reflect the saved state, so they should logically be grouped together in some way.

app/src/main/java/mozilla/lockbox/support/Optional.kt
sashei: The concept of optionals in our observables is useful; we use the presence / absence of things like OAuthInfo to do routing in iOS and I'm trying to think of how it would look otherwise:

We could be more _explicit_ by having something like `val accountAvailable: Observable<Bool>` but it's another piece of state to update along with `oauthInfo` / `profile`. 

To avoid updating that state every time, we could just implement a filter for `oauthInfo == null`, which brings us back to the original problem of not being able to represent Optionals in Rx. I think we'll want them in other places so I'm OK with this solution for now.

At its core, this is a workaround for RxKotlin being implemented in Java and Java not having a concept of Optionals.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
sashei: I also think of `load` and `persist` as extending to loading / persisting things into our external `Observable`s.

app/src/test/java/mozilla/lockbox/support/SecurePreferencesTest.kt
```diff
@@ -0,0 +1,163 @@
...
+import android.content.SharedPreferences
+import android.util.Base64
+import mozilla.components.lib.dataprotect.Keystore
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.Mockito
+import org.mockito.Mockito.`when` as whenCalled
```
jhugman: Ooh, fancy `as`.

app/src/test/java/mozilla/lockbox/store/AccountStoreTest.kt
```diff
@@ -0,0 +1,33 @@
...
+//
+//    @Test
+//    fun `it pushes null when there is no saved fxa information`() {
+//        verify(securePreferences.getString("firefox-account"))
+//
+//        oauthObserver.assertValue(Optional<OAuthInfo>(null))
+//        profileObserver.assertValue(Optional<Profile>(null))
+//    }
+
+    // note: further FxA-related tests on hold until we can use x-compiled Rust code in unit specs
```
jhugman: Ok.

app/src/main/java/mozilla/lockbox/support/SecurePreferences.kt
jhugman: 👍 

Eventually, I think we should try and make this a version for `android-components` that is a drop in replacement/wrapper for `SharedPreferences`.

app/src/test/java/mozilla/lockbox/ExampleUnitTest.kt
```diff
@@ -1,23 +0,0 @@
...
- * Example local unit test, which will execute on the development machine (host).
- *
- * See [testing documentation](http://d.android.com/tools/testing).
- */
-class ExampleUnitTest {
-    @Test
-    fun addition_isCorrect() {
-        assertEquals(4, 2 + 2)
-    }
-}
```
jhugman: 👍 All the thumbs. I hadn't realised this was still here.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -0,0 +1,80 @@
...
+            profileSubject.onNext(it.asOptional())
+        }
+
+        val oauthSubject = oauthInfo as PublishSubject
+        account.getOAuthToken(FXA_SCOPES).whenComplete {
+            oauthSubject.onNext(it.asOptional())
+        }
+    }
+
+    private fun pushNullAndReset() {
```
jhugman: Note: I hate names with conjunctions in. Please don't fix this in this PR.

PR #156, opened 2018-10-18
Rearrange the navigation graph to ensure the lockscreen clears the back stack

### LINKED ISSUE #143: Ensure Lock screen clears the back stack
As a followup to PR #133, this story ensures that no portion of the app is navigatable to from the lock screen, except by pressing the "Unlock" button — or if it has been implemented, by unlocking it.

This may mean separating the lockscreen off into its own navigation graph, or activity.

Acceptance criteria: 

 * [ ] The entry list is not accessible via the back button.
 * [ ] The lock screen is not accessible from the entry list (or the screen from whence it became locked) after unlocking and pressing the back button.

jhugman:
Fixes #143

This PR re-implements detecting the menu item selections to use the dispatcher to call the `navController`.

This is because the blessed way of using the appDrawer with the navController constructed its own action (and animations) instead of using the one we specifiy in the `graph.xml`. This adds those default animations to the settings action, and changes the action to the locked screen to clear the back stack.

Additionally, the locked screen is moved to its own subgraph.

Unfortunately, the action to transition to the locked screen can't be made into a global action yet, due to a bug in the navigation library.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
@@ -6,21 +6,25 @@
 
 package mozilla.lockbox.presenter
 
+import android.support.annotation.IdRes
 import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.rxkotlin.addTo
+import mozilla.lockbox.R
 import mozilla.lockbox.action.DataStoreAction
 import mozilla.lockbox.action.RouteAction
 import mozilla.lockbox.extensions.mapToItemViewModelList
 import mozilla.lockbox.flux.Dispatcher
 
 import mozilla.lockbox.flux.Presenter
+import mozilla.lockbox.log
 import mozilla.lockbox.model.ItemViewModel
 import mozilla.lockbox.store.DataStore
 
 interface ItemListView {
     val itemSelection: Observable<ItemViewModel>
     val filterClicks: Observable<Unit>
+    val menuItemSelections: Observable<Int>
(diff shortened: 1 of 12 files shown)


app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
```diff
@@ -49,7 +53,21 @@ class ItemListPresenter(
...
+            .subscribe(this::onMenuItem)
+            .addTo(compositeDisposable)
+
         // TODO: remove this when we have proper locking / unlocking
         dispatcher.dispatch(DataStoreAction.Unlock)
     }
+
+    fun onMenuItem(@IdRes item: Int) {
+        val action = when (item) {
+            R.id.action_itemList_to_locked, R.id.fragment_locked -> RouteAction.LockScreen
```
sashei: in what situation would the `menuItemSelection` return an `R.id.action_...`? unfamiliar with exactly how the `nav_menu` fits in here (more a personal question than review issue)

app/src/main/res/navigation/graph_locked.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sashei: I like the `graph_...` naming convention

app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
jhugman: Currently, there is no merit in having this accept both the action and the destination. 

I'm hard pressed to justify this, except the vague feeling we should move the the action to be a global action. On the whole, that should be invisible to this code.

PR #174, opened 2018-10-29
Implement Pattern/PIN/Password Fallback

### LINKED ISSUE #69: Implement Pattern/PIN/Password Fallback
Implement a fallback to use the Pattern/PIN/Password configured on the user's device when:

* the user selects "USE DEVICE PIN" from the app's biometric prompt
* the user does not have any fingerprints enrolled
* the user's device does not have nay fingerprint reading hardware

Use the KeyguardManager to prepare an Intent for this prompt, and respond appropriately to `ACTIVITY_OK` (unlock) or `ACTIVITY_CANCELED` (stay locked).

ioana-farcas:
Fixes #69 

 ## Testing and Review Notes
Tested on Pixel 2 phone, Api level 27.
@nickbrandt The screen that is displayed when falling back to Pattern/PIN or Password is not a custom screen, it is launched using an intent that receives only a **title** and a **description**. Because of this, there is no way to control where and how the error and other items are displayed (this screen could look different on different devices from different vendors). Also, there is no way to identify if the "credentials" are a pin, pattern or password. I would suggest to have a more general description to pass to the activity, something like: "Confirm your screen lock PIN, Pattern or Password". Please take a look and let me know what do you think

 ## To Do
 - Add “WIP” to the PR title if pushing up but not complete nor ready for review
- [ ] add unit tests


nickbrandt:
@ioana-farcas yes, that makes sense. Please see the revised copy in Zeplin: https://zpl.io/bAqXRm6

We understand the look will change depending on the device. Thank you!

ioana-farcas:
@nickbrandt I added the change for the error text. Please take a look. Thanks!

nickbrandt:
@ioana-farcas looks good, thank you!

app/src/main/java/mozilla/lockbox/store/LockedStore.kt
```diff
@@ -14,11 +15,19 @@ import mozilla.lockbox.flux.Dispatcher
...
 
     open val onAuthentication: Observable<FingerprintAuthAction> =
         dispatcher.register
             .filterByType(FingerprintAuthAction::class.java)
+
+    fun apply(manager: KeyguardManager) {
+        keyguardManager = manager
+    }
+
+    val isKeyguardSecure get() = keyguardManager.isKeyguardSecure
```
sashei: Rather than having this in the `LockedStore`, I'd put this in the `FingerprintStore` where we are checking other device security settings (corresponds to @linuxwolf 's desire to have all device security information in one place)

It means passing two things to the `FingerprintStore.apply()` method which is 🤮 but will get fixed up by #177

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -46,13 +48,16 @@ class LockboxApplication : Application() {
...
         FingerprintStore.shared.apply(getSystemService(Context.FINGERPRINT_SERVICE) as FingerprintManager)
+        LockedStore.shared.apply(getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager)
 
         // use context for PreferenceManager
         SecurePreferences.shared.apply(PreferenceManager.getDefaultSharedPreferences(this))
 
         // hook this context into Telemetry
         TelemetryStore.shared.applyContext(this)
 
+        FingerprintStore.shared.applyContext(this)
```
sashei: can we set up our `lateinit` properties (such as `FingerprintManager`, `KeyguardManager`) in this `applyContext` method instead of instantiating them above?

sashei: ^ that would obviate the need for the `apply` method in the `FingerprintStore` which would be awesome

ioana-farcas:
@sashei I added the requested changes, thanks for the review :)

PR #196, opened 2018-11-08
Codecov tweaks

### LINKED ISSUE #187: Review codecov configuration
via quick chat with @sashei since some stuff we can't really test:

- UI / views
- selectively for espresso vs unit
- FxA (will be low)
- get yaml file included in the codebase

devinreams:
Fixes #187

## Testing and Review Notes

Make sure code cov is better than 50% now 😉

### DIFF
--- a/app/src/main/java/mozilla/lockbox/extensions/view/AlertDialog+.kt
+++ b/app/src/main/java/mozilla/lockbox/extensions/view/AlertDialog+.kt
@@ -4,12 +4,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-/*
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
- */
-
 package mozilla.lockbox.extensions.view
 
 import android.content.Context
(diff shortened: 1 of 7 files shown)


app/src/main/java/mozilla/lockbox/extensions/view/AlertDialog+.kt
```diff
@@ -4,7 +4,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package mozilla.lockbox.extensions
+/*
```
devinreams: Whoops, extra blocks snuck in

devinreams:
OK looks like we're closer to 64% overall if we exclude the view stuff:

screen shot 2018-11-08 at 10 43 00 am

https://codecov.io/gh/mozilla-lockbox/lockbox-android/tree/2bdec8b05bf1a615aa44169412c9592f0a098ad6/app/src/main/java/mozilla/lockbox

Thanks for looking over my shoulder, @sashei. 🙇

devinreams:
And here's what it looks like in PR status.

image


Which gives me an idea to try splitting just the AccountStore into its own reported number and then "everything else". 🤔

devinreams:
OK I'm done playing with it. Here two new statuses that show:

- overall project (project) = 64%
- AccountStore coverage (project/fxa) = 5%
- everything but that ☝️ (project/app) = 66%

screen shot 2018-11-08 at 11 28 11 am

PR #218, opened 2018-11-16
194 fingerprint dialog orientation changed

### LINKED ISSUE #194: [Crash] App crashes on orientation changed, when fingerprint dialog is displayed
Go to Lock screen or Settings screen. Tap on Unlock button or toggle button for enabling fingerprint. After the fingerprint dialog is displayed, change the orientation of the device => app crashes

ioana-farcas:
Fixes #194 

## Testing and Review Notes
Please verify behaviour at orientation change for Lock screen and Settings screen.

## To Do

- [x] add unit tests
- [ ] ~~check on the [accessibility](https://mozilla-lockbox.github.io/lockbox-android/accessibility/) of any added UI~~

### DIFF
--- a/app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
+++ b/app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
@@ -53,7 +53,7 @@ class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
     }
 
     override fun onBindViewHolder(holder: SettingViewHolder, position: Int) {
-        holder.disposable?.let { compositeDisposable.remove(holder.disposable!!) }
+        holder.disposable?.let { compositeDisposable.remove(it) }
 
         val configuration = settingListConfig[position]
         when {
--- a/app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
+++ b/app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
@@ -9,6 +9,7 @@ package mozilla.lockbox.adapter
 import android.support.v7.widget.RecyclerView
 import android.view.LayoutInflater
 import android.view.ViewGroup
+import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.addTo
 import mozilla.lockbox.R
 import mozilla.lockbox.view.AppVersionSettingViewHolder
@@ -18,6 +19,7 @@ import mozilla.lockbox.view.ToggleSettingViewHolder
 
 class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
     private var settingListConfig: List<SettingCellConfiguration> = emptyList()
+    private val compositeDisposable = CompositeDisposable()
 
     companion object {
         const val SETTING_TEXT_TYPE = 0
@@ -51,7 +53,7 @@ class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
     }
 
     override fun onBindViewHolder(holder: SettingViewHolder, position: Int) {
-        holder.compositeDisposable.clear()
+        holder.disposable?.let { compositeDisposable.remove(holder.disposable!!) }
 
         val configuration = settingListConfig[position]
         when {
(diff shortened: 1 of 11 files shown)


app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
```diff
@@ -81,24 +83,20 @@ class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
...
+            else -> throw IllegalStateException("Please use a valid defined setting type.")
         }
     }
 
     fun setItems(settingConfig: List<SettingCellConfiguration>) {
         settingListConfig = settingConfig
         notifyDataSetChanged()
     }
+
+    override fun onDetachedFromRecyclerView(recyclerView: RecyclerView) {
```
sashei: ooo

app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
```diff
@@ -81,24 +83,20 @@ class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
...
-                return SETTING_TOGGLE_TYPE
-            }
-            is AppVersionSettingConfiguration -> {
-                return SETTING_APP_VERSION_TYPE
-            }
-            else -> {
-                throw IllegalStateException("Please use a valid defined setting type.")
-            }
+        return when (configurationType) {
+            is TextSettingConfiguration -> SETTING_TEXT_TYPE
```
sashei: 👍

app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
```diff
@@ -51,7 +53,7 @@ class SettingListAdapter : RecyclerView.Adapter<SettingViewHolder>() {
     }
 
     override fun onBindViewHolder(holder: SettingViewHolder, position: Int) {
-        holder.compositeDisposable.clear()
+        holder.disposable?.let { compositeDisposable.remove(holder.disposable!!) }
```
jhugman: This looks needlessly complicated. Also avoid `!!`.

Would work: 
```kt
 holder.disposable?.let { compositeDisposable.remove(it) }
```

app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
jhugman: 👍

app/src/main/java/mozilla/lockbox/presenter/SettingPresenter.kt
```diff
@@ -43,15 +44,18 @@ class SettingPresenter(
 
     private val versionNumber = BuildConfig.VERSION_NAME
 
-    private val autoLockObserver: Consumer<Boolean>
+    val autoLockObserver: Consumer<Boolean>
```
jhugman: I'm confused why this is a consumer, and everything else is an `Observer`, or a closure used to subscribe.

A tiny comment here would be helpful. 

```
// we use a Consumer here because…
```

app/src/main/java/mozilla/lockbox/adapter/SettingListAdapter.kt
ioana-farcas: 👍

app/src/main/java/mozilla/lockbox/presenter/SettingPresenter.kt
ioana-farcas: For each toggle button a Consumer is used as a callback in the subscribe method. This is because when switching the toggle button, there is no need for the onError or other events except onNext . As far as I understand, when only one event is needed, there is no reason to use an Observer and implement all the events (onNext, onError, onComplete..). Instead a Consumer can be used. I will add a comment if you think it's necessary.

sashei: IMO the comment might not be necessary; we only use `Consumer`s in the `SettingPresenter`. It also has the advantage of tying the lifecycle to the view.

jhugman: :+1:

ioana-farcas:
@jhugman I changed the naming to **UnlockWithFingerprintPendingAuth**. Please let me know if this is ok with you

Issue #55, opened 2018-07-12
Enable autofill (onboarding)


sandysage:
## Acceptance criteria
* Should be able to set up autofill settings (OS) from the Lockbox onboarding flow

### Engineering Notes

On iOS, this required "pausing" in driving the UI from the datastore state while onboarding. We added a `val onboarding: Observable<Boolean>` to the `RouteStore`, and ignored incoming non-onboarding routes while it was true.

Make sure that navigating to this screen prevents the user from seeing the ugly webpage error as referenced in #273

### Visual Design

**InVision**
Onboarding screen: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/309979871_AutoFill
Opens Device Autofill Settings: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/314387970_AutoFill_-_Settings
Confirmation Dialog: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/314387969_AutoFill_-_Settings_-_Confirm

**Zeplin**
Onboarding screen: https://zpl.io/agNkEy0
Opens Device Autofill Settings: https://zpl.io/VYRpzym
Confirmation Dialog: https://zpl.io/agNkEXN

Depends on #214
Depends on #215
Depends on #216

Issue #35, opened 2018-07-12
If device pin/biometrics not set, provide instructions to do so (onboarding)


sandysage:
We need to communicate and support users to set this up without requiring device authentication to use the app.

## Requirements
* display the passcode prompt (when appropriate) on `Get Started` button tap in the welcome page
* if the user taps `Skip`, just route to FxA
* if the user hits set passcode, route to the settings page and remain on the welcome screen in the application
* don't bother storing the `Skip` preference, just launch it (when warranted) every time the user hits `Get Started`

### Engineering Notes

On iOS, this required "pausing" in driving the UI from the datastore state while onboarding. We added a `val onboarding: Observable<Boolean>` to the `RouteStore`, and ignored incoming non-onboarding routes while it was true.

Make sure that navigating to this screen prevents the user from seeing the ugly webpage error as referenced in #273

### Visual Design

**InVision**
https://mozilla.invisionapp.com/share/3SN35UHWJDX#/312603009_No_PIN

**Zeplin**
https://zpl.io/agvL3GN

Depends on #70
Depends on #30

Issue #30, opened 2018-07-12
Enable biometric unlock (onboarding)


sandysage:
## Acceptance criteria
* Should provide a way to skip this step if user chooses not to enable
* On choosing to enable, user should be prompted to authenticate with biometrics
* Only display this to users that: 
  1. Have Fingerprint available on their device  
  2. Have this feature enabled on their device (have a Fingerprint set).  

Make sure that navigating to this screen prevents the user from seeing the ugly webpage error as referenced in #273

requires #28 

### Engineering Notes

On iOS, this required "pausing" in driving the UI from the datastore state while onboarding. We added a `val onboarding: Observable<Boolean>` to the `RouteStore`, and ignored incoming non-onboarding routes while it was true.

### Visual Design

**InVision**
Initial: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/309979870_Fingerprint
Error: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/310610373_Fingerprint_-_Error
Success*: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/311048434_Fingerprint_-_Success
*Intent would be that the success message displays, pauses for a second, then transitions to next screen in onboarding (AutoFIll)

**Zeplin**
Initial: https://zpl.io/VYBxqRk
Error: https://zpl.io/Vk5l1nW
Succes: https://zpl.io/25qK4rW

Issue #24, opened 2018-07-12
Onboarding confirmation with security theatre (onboarding)


sandysage:
Because mobile access is complementary to the browser experience, I want to retrieve logins that I have already saved, and need to understand what is required to do that (having saved logins in Firefox, FxA + Sync enabled)

## Acceptance criteria
* Should provide instruction to get a Firefox account & enable Sync
* Should provide context of what saved logins are in the browser, in case none are saved
* Content should provide context around safety and security

Make sure that navigating to this screen prevents the user from seeing the ugly webpage error as referenced in #273

### Visual Design

**InVision**
https://mozilla.invisionapp.com/share/3SN35UHWJDX#/309979872_Onboarding_Confirmation

**Zeplin**
https://zpl.io/V0yxoJQ

Depends on #30

PR #232, opened 2018-11-23
Loading indication for sync

### LINKED ISSUE #48: Loading Indication for Sync
When loading up my entries for the very first time, I should be given an affordance of something happening within the app, e.g., a loading indication, if it takes more than two seconds to load and import the data into the app.

## Acceptance criteria
* Should default display loading screen everytime
* Should not be able to perform any list actions such as search or sort during this interaction
* Should be able to navigate to Settings even when loading indication appears

### Visual Design

**InVision**
https://mozilla.invisionapp.com/share/3SN35UHWJDX#/310347758_Entry_List_-_Syncing

**Zeplin**
https://zpl.io/Vk5l1KL

Requires #43
Depends on #244

ioana-farcas:
Fixes #48 
## Testing and Review Notes
@mozilla-lockbox/ux 
I can not tell if the bottom bar with the spinner has a shadow in zeplin...so I did not add one. Please let me know if it is needed.

## Screenshots or Videos

loading

## To Do

- [x] add unit tests


devinreams:
Some refactor required per @ioana-farcas and @jhugman...

sashei:
@changecourse a few things:

- the 404 page is expected for now, we aren't navigating away from the FxA redirect quickly enough for the time being and the expectation is that this will be fixed with the implementation of the onboarding stories (all onboarding stories reflect this expectation #30 #24 #35 #55)
- I have experienced similar issues switching from using test data to real data, I have to do a hard uninstall-reinstall cycle every time, so I don't think this is related to this PR necessarily. I haven't filed an issue for that but can do that today if others are having issues!!

changecourse:
@sashei thanks for the clarifications! As long as we're tracking both of those, we're good then :)

UI-wise, we're good.

devinreams:
>I have experienced similar issues switching from using test data to real data, I have to do a hard uninstall-reinstall cycle every time, so I don't think this is related to this PR necessarily. I haven't filed an issue for that but can do that today if others are having issues!!

Ah interesting, I haven't seen that myself but since I was here already, went ahead and filed that issue as a P2: #277

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -48,7 +48,7 @@ open class DataStore(
     private val listSubject: BehaviorRelay<List<ServerPassword>> = BehaviorRelay.createDefault(emptyList())
 
     val state: Observable<State> get() = stateSubject
-    val syncState: Observable<SyncState> = PublishSubject.create()
+    open val syncState: Observable<SyncState> = ReplaySubject.create()
```
sashei: 👍

app/src/main/java/mozilla/lockbox/view/ItemListFragment.kt
```diff
@@ -202,4 +203,15 @@ class ItemListFragment : CommonFragment(), ItemListView {
...
-}
\ No newline at end of file
+
+    override fun loading(isLoading: Boolean) {
+        if (isLoading) {
+            showAndRemove(view!!.loadingView, view!!.entriesView)
+        } else {
+            showAndRemove(view!!.entriesView, view!!.loadingView)
+        }
+        view!!.filterButton.isClickable = !isLoading
```
sashei: This could be a follow-up improvement, but it would be nice for these `isClickable` and `isEnabled` settings to each be their own `Consumer`s on the `View` rather than side effects of showing / hiding the loading spinner. IIRC from iOS, we have other situations (no saved entries comes to mind) where we want to disable the filter / sort buttons. However, I'm happy to let this PR merge as-is and do that work in later tickets :)

ioana-farcas: @sashei You're right. I will merge this and add the modifications in another PR. Thanks for the review!

devinreams: Filed a follow-up issue here: https://github.com/mozilla-lockbox/lockbox-android/issues/279

Issue #32, opened 2018-07-12
Biometric Access Timer


sandysage:
## Acceptance criteria
* Should be able to set a timer to lock saved logins
* Should default to being on
* Should be able to set via the settings page (onboarding is different story)
* Should be able to unlock from the same UI as signing in for the first time (until there is a use case for those to be different)
* Should default to 5 minutes until the user sets the timer*
* Should _**not**_ take effect while in the foreground
* Should start when moved to the background
* Should start when quit

### Visual Design

**InVision**
Auto lock dialog: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/309979888_Settings_-_Auto_Lock

**Zeplin**
Auto lock dialog: https://zpl.io/bLOR9BJ

### Engineering Considerations

* Use absolute UTC date/times for the autolock time
* Calculate and set the autolock time when:
  * the app first starts
  * the app goes into the background (including when the screen goes to sleep)
  * the autolock setting is changed
  * the app transitions to a new activity/fragment (e.g., list view to detail view)
* Assume the app is in the locked state until the autolock time can be compared to the current time

sashei:
@linuxwolf @changecourse @nickbrandt @sandysage did we ever come to a decision about not locking while the app is in the foreground, and instead treating the autolocktime interval as something that "starts" when the app goes into the background?

linuxwolf:
> @linuxwolf @changecourse @nickbrandt @sandysage did we ever come to a decision about not locking while the app is in the foreground, and instead treating the autolocktime interval as something that "starts" when the app goes into the background?

We haven't as far as I know.  I think it's something we should do, but we haven't discussed it.  I'll get an email with pros/cons (as I see them), and we can discuss quickly.

sashei:
Split out issue #219 for the security disclaimer to limit scope on this issue.

PR #240, opened 2018-11-27
autolocking

### LINKED ISSUE #234: Make AutolockTime setting lock the app
This is a follow-up to #32.

Right now, users can set an Autolock time but it will not do anything w.r.t. locking the app.

This ticket will integrate the autolocktime set by the user to lock the app at appropriate times.

#### Engineering considerations:
- Use absolute UTC date/times for the autolock time
- Calculate and set the autolock time when:
  - the app first starts
  - the app goes into the background (including when the screen goes to sleep)
  - ~~the autolock setting is changed~~
  - ~~the app transitions to a new activity/fragment (e.g., list view to detail view)~~
- Assume the app is in the locked state until the autolock time can be compared to the current time

engineering tasks:
- [x] read & propagate autolocktime setting on foreground
- [x] lock datastore when directed
- [x] write to autolocktime for relevant actions
- [x] check `boot_id` for lock status
- [x] decide on securepreferences for storage?
- [x] handle Lock Now
- [x] functional testing
- [x] handle backgrounding / foregrounding during login flow (or from welcome screen)
- [ ] adjust UI tests

depends on #144
depends on #275

sashei:
Fixes #234
Connected to #32

## Testing and Review Notes
the autolock time reset is only triggered on _backgrounding_ the app including hard-quit.

##### backgrounding only:
- set timer to 1 minute, background for longer than 1 minute, open app, locked ✅ 
- set timer to 1 minute, background for <1 minute, open app, unlocked ✅ 
- set timer to 15 minutes, background for <15 minutes, open app, unlocked ✅ 
- set timer to 15 minutes, tap "Lock Now", background for <15 minutes, open app, locked ✅ 
- used "Lock Now", backgrounded app, returned, still locked ✅ 

##### hard-quitting tests:
- set timer to 1 minute, hard-quit and wait for longer than 1 minute, open app, locked ✅ 
- set timer to 1 minute, hard-quit and wait for <1 minute, open app, unlocked ✅ 
- set timer to 15 minutes, hard-quit and wait for <15 minutes, open app, unlocked ✅ 
- set timer to 15 minutes, tap "Lock Now", hard-quit and wait for <15 minutes, open app, locked ✅
- used "Lock Now", hard-quit app, returned, still locked ✅

##### login tests
- start app, go to fxa login, background for < 5 minutes, open app, still on FxA ✅ 
- start app, go to fxa login, background for > 5 minutes, open app, still on FxA ✅ 
- start app, go to fxa login, hardquit & wait for < 5 minutes, open app, on welcome screen ✅ 
- start app, go to fxa login, hardquit & wait for > 5 minutes, open app, on welcome screen ✅ 
- disconnect account, go to fxa login, background for < 5 minutes, open app, still on FxA ✅ 
- disconnect account, go to fxa login, background for > 5 minutes, open app, still on FxA ✅ 
- disconnect account, go to fxa login, hardquit & wait for < 5 minutes, open app, on welcome screen ✅ 
- disconnect account, go to fxa login, hardquit & wait for > 5 minutes, open app, on welcome screen ✅ 

## To Do

- [x] add unit tests


app/src/main/java/mozilla/lockbox/action/SettingAction.kt
```diff
@@ -44,6 +46,24 @@ class Setting {
...
+                FifteenMinutes -> 60 * 15
+                ThirtyMinutes -> 60 * 30
+                OneHour -> 60 * 60
+                TwelveHours -> 60 * 60 * 12
+                TwentyFourHours -> 60 * 60 * 24
+                Never -> 0
+            }
+        }
+
+        val ms: Long by lazy { (this.seconds * 1000).toLong() }
```
jhugman: Enums in kotlin can have their own constructors and properties.

You can get rid of your Swift-y switches :); for example:

e.g.
```kotlin
enum AutoLockTime(
  val stringValue: @StringRes Int,
  private val minutes: Long
) {
    val second: Long = minutes * 60L
    val ms: Long = seconds * 1000L

    OneMinute(R.string.one_minute, 1),
    TwoMinute(R.string.one_minute, 2),
    …
}
```

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,51 @@
...
+    val lifecycleStore: LifecycleStore = LifecycleStore.shared
+) : ContextStore {
+
+    private val compositeDisposable = CompositeDisposable()
+    private lateinit var preferences: SharedPreferences
+
+    init {
+        lifecycleStore.lifecycleFilter
+            .filter { it == LifecycleAction.Background }
+            .switchMap { settingStore.autoLockTime.take(1) }
```
jhugman: Mind. Blown.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,51 @@
...
+                val autoLockTimerDate = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    Instant.now().plusMillis(it.ms).toEpochMilli()
+                } else {
+                    TODO("VERSION.SDK_INT < O")
+                }
+
+                preferences
+                    .edit()
+                    .putLong(Constant.Key.autoLockTimerDate, autoLockTimerDate)
+                    .apply()
```
jhugman: Putting `autoLockTimerDate` and `autoLockTime` in `preferences` — rather than `SecurePreferences` — makes me a little nervous.

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
```diff
@@ -44,6 +46,24 @@ class Setting {
                     Never -> R.string.never
                 }
             }
+
+        val seconds: Int by lazy {
+            // I'm concerned about this method of setting time. I wonder if it's going
+            // to be circumventable by changing the device time. we might want to consider a
+            // options for mitigating that attack vector.
```
jhugman: We should probably use [`SystemClock.elapsedRealtime()`][0] instead of translating to a date-time. If we are thinking about `AlarmManager`, then we should double check that the calculate `elapsedRealtime` actually has passed.

I'm not sure what the status of the designs are, but we should definitely force the user to unlock after a reboot.

[0]: https://developer.android.com/reference/android/os/SystemClock.html

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
sashei: oooo

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
sashei: @linuxwolf do you have an opinion about this?

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
sashei: +1. Have explored the `SystemClock.elapsedRealtime()` option and I think that's the way to go.

@changecourse @nickbrandt are y'all OK with the app locking if the device has rebooted, regardless of time elapsed?

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private fun checkElapsedTime(autoLockTime: Setting.AutoLockTime) {
+        val lastBackgroundTime = preferences.getLong(Constant.Key.backgroundedTime, -1)
+        val currentSystemTime = SystemClock.elapsedRealtime()
+
+        val elapsedMS = currentSystemTime - lastBackgroundTime
+
+        when {
+            // edge case: someone has backgrounded the app after the device has been running for x minutes.
```
sashei: @jhugman would appreciate thoughts on strategy here -- should the store dispatch the lock / unlock actions?

in iOS we basically have the datastore read the stored `backgroundedTime` equivalent and decide whether it is locked or unlocked IIRC, but it would be nice to have it a more contained to this `AutoLockStore` class.

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
nickbrandt: @sashei yea that makes sense to me 👍

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
linuxwolf: Throwing everything into `SecurePreferences` is an anti-pattern not to be repeated.

Things that go through `SecurePreferenecs` are ones where an observer can do something with them later -- such as the encryption keys or the user's information.  Settings like this do not have that concern, so shouldn't go into `SecurePreferences`.

app/src/test/java/mozilla/lockbox/store/DataStoreTest.kt
```diff
@@ -86,7 +86,7 @@ class DataStoreTest : DisposingTest() {
     @Test
     fun testTouch() {
         dispatcher.dispatch(DataStoreAction.Unlock)
-        sleep(100)
+        sleep(200)
```
linuxwolf: ugh

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
```diff
@@ -20,30 +20,20 @@ open class SettingAction : Action {
...
+        private val seconds: Long
+    ) {
+        OneMinute(R.string.one_minute, 60),
+        FiveMinutes(R.string.five_minutes, 60 * 5),
+        FifteenMinutes(R.string.fifteen_minutes, 60 * 15),
+        ThirtyMinutes(R.string.thirty_minutes, 60 * 30),
+        OneHour(R.string.one_hour, 60 * 60),
+        TwelveHours(R.string.twelve_hours, 60 * 60 * 12),
+        TwentyFourHours(R.string.twenty_four_hours, 60 * 60 * 24),
+        Never(R.string.never, 0);
```
linuxwolf: Assuming we eventually support "Lock Immediately", what might its `seconds` be?  `-1`?

I don't know that it's worth even considering right now, but I hope eventually 😄

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private fun checkElapsedTime() {
+        val autoLockTimerDate = preferences.getLong(Constant.Key.autoLockTimerDate, -1)
+        val currentSystemTime = SystemClock.elapsedRealtime()
+
+        // edge case: someone has backgrounded the app after the device has been running for x minutes.
+        // they reboot the device and leave it running for x minutes + some time less than autolocktime setting
+        // app will be unlocked :(
+        val diff = autoLockTimerDate - currentSystemTime
```
linuxwolf: This is a little concerning to me, but not sure if we can properly address it.

I wonder if adding a broadcast receiver for system boot is worth exploring.  My initial thought is to reset the `autolockTimerDate` to 0 (or delete it altogether).  I _think_ that would lock by default.

Thoughts? (cc @jhugman)

app/src/main/res/values/strings.xml
```diff
@@ -91,7 +91,6 @@
     <string name="sort_description">Sort entries</string>
     <string name="search_clear_content_description">Clear search bar text</string>
     <string name="profile_image_content_description">Firefox Account Profile Image</string>
-    <string name="app_version">Application version number</string>
```
linuxwolf: Is this intentional for this PR?

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
sashei: tbh I don't think it would be that hard; we'd just lock on every foreground without checking the stored time at all if that was the user setting :p

app/src/main/res/values/strings.xml
sashei: just some cleanup when I was fixing the baseline lint file 🛀

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
linuxwolf: touché (-:

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
sashei: I've done a little investigating; we can either implement a broadcast receiver, or (IMO best solution), store the value found at `/proc/sys/kernel/random/boot_id` and check it on startup. This id is re-generated with each boot, and if it's different than our stored value, we lock the app.

app/src/androidTest/java/mozilla/lockbox/RoutePresenterTest.kt
```diff
@@ -48,19 +48,22 @@ open class RoutePresenterTest {
 
     @Test
     fun testItemList() {
+        Dispatcher.shared.dispatch(DataStoreAction.Unlock)
```
jhugman: Perhaps we should stick all these in a `navigator` method?

app/src/main/java/mozilla/lockbox/action/SettingAction.kt
```diff
@@ -20,30 +20,20 @@ open class SettingAction : Action {
...
-                    FiveMinutes -> R.string.five_minutes
-                    FifteenMinutes -> R.string.fifteen_minutes
-                    ThirtyMinutes -> R.string.thirty_minutes
-                    OneHour -> R.string.one_hour
-                    TwelveHours -> R.string.twelve_hours
-                    TwentyFourHours -> R.string.twenty_four_hours
-                    Never -> R.string.never
-                }
-            }
+        val ms: Long by lazy { (this.seconds * 1000) }
```
jhugman: Nit: probably a little lightweight for a `lazy` property.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
jhugman: @linuxwolf I don't understand this reasoning.

I'm working on the assumption that `SharedPreferences` is not good a good place for secrets to be put — we wish to prevent unauthorized reads. And that `autolockDateTime` should not be written by anyone except us, lest it be written as an expired time — we wish to prevent unauthorized writes.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,73 @@
...
+        lifecycleStore.lifecycleFilter
+            .filter { it == LifecycleAction.Background }
+            .switchMap { settingStore.autoLockTime.take(1) }
+            .subscribe {
+                val currentSystemTime = SystemClock.elapsedRealtime()
+                preferences
+                    .edit()
+                    .putLong(Constant.Key.autoLockTimerDate, currentSystemTime + it.ms)
+                    .apply()
+            }
```
jhugman: When to store the `boot_id` to compare here may end up being problematic, though I'd guess it were here.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
sashei: in our implementation, `SecurePreferences` is just an encryption-layer wrapper for `SharedPreferences`, so we shouldn't have the same issues clearing it that we do on iOS.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
linuxwolf: sounds good to me!  The broadcast receiver was the first thing I found about detecting reboot.  As long as `boot_id` exists since API 23 and it can be read without extra permissions, let's do it.

linuxwolf:
> I'm working on the assumption that SharedPreferences is not good a good place for secrets to be put — we wish to prevent unauthorized reads. And that autolockDateTime should not be written by anyone except us, lest it be written as an expired time — we wish to prevent unauthorized writes.

Encrypting does not prevent unauthorized writes.  It might mean we reject the value if it can't be decrypted, but a write also means the value could be completely deleted and then have a similar effect.  It's also not clear what exactly the attack this prevents and is worth trying to protect against.

Exfiltration of user information is a well-understood attack that we can definitely do something about.  The value of user's personal (email, displayname, avatar, uid) and sensitive (encryption keys, credentials) information is high value; the value of `autolockTimerDate` and other settings is not.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
linuxwolf: [ moved to keep the threading ]

> I'm working on the assumption that SharedPreferences is not good a good place for secrets to be put — we wish to prevent unauthorized reads. And that autolockDateTime should not be written by anyone except us, lest it be written as an expired time — we wish to prevent unauthorized writes.

As @sashei indicates, `SecurePreferenes` is a wrapper over `SharedPreferences`.  This means taht encrypting does not prevent unauthorized writes.  It means we might reject the value if it can't be decrypted, but then the item could be deleted instead and have a similar effect.  It's also not clear what exactly the attack this prevents and is worth trying to protect against.

Exfiltration of user information is a well-understood attack that we can definitely do something about.  The value of a user's personal (email, displayname, avatar, uid) and sensitive (encryption keys, credentials) information is high value; the value of `autolockTimerDate` and other settings is not.

sashei: Here's my possible attack vector (@linuxwolf lmk what you think): someone could write a later datetime to the `autolocktime` shard preference if they knew how we were going to be reading it. We would then read the later date time as the next lock time, see that the system clock is behind that, and unlock. At least behind the encryption layer, the value will get decrypted to garbage and we can just lock when that happens.

linuxwolf: Ok, and the attacker does what with that?  What does the attack actually do?  Does it leak data?  Does it give the attack access to the database?

This is the problem I have with these.  I'm all for protecting data, but let's be smart about what we are protecting and why.  Protecting encryption keys and user info means an attacker can't then use that to get real data.

sashei:
@linuxwolf @sashei @eliserichards let me know what you think about the updated *testing note* for this issue...

linuxwolf:
> let me know what you think about the updated testing note for this issue...

I'm trying to remember the rationale for updating the timer on route actions; my original proposal was to set the timer on backgrounding (to align with suspected user expectations) and startup (to make sure we don't start unlocked if the app crashes instead of backgrounding).

This change to only on route actions may very well have unexpected results for a user.  I can see a case where someone has the item detail open and foregrounds/backgrounds a couple of times.  Personally, I'd rather ditch updates on route transitions than backgrounding.

sashei:
@linuxwolf agreed, I'm amending the requirements on the original issue #234 to remove route actions (and setting a different autolocktime??) as triggers for resetting the `autolocktimerdate`. that ameliorates some wiring complexity, too.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,111 @@
...
+        return true
+    }
+
+    private fun autoLockTimePassed(): Boolean {
+        val autoLockTimerDate = preferences.getLong(Constant.Key.autoLockTimerDate, -1)
+        val currentSystemTime = SystemClock.elapsedRealtime()
+
+        val diff = autoLockTimerDate - currentSystemTime
+
+        return diff <= 0
```
jhugman:
```kotlin
return autoLockTimerDate <= currentSystemTime
```

app/src/main/java/mozilla/lockbox/support/FileReader.kt
```diff
@@ -0,0 +1,10 @@
+package mozilla.lockbox.support
+
+import java.io.FileReader
+
+class FileReader {
+    fun readContents(filePath: String): String {
+        val file = FileReader(filePath)
+        return file.readText()
```
jhugman: Erm. Ok. I'm not sure I understand why so much code here.

But we're not passing this into the store, so we can test it?

app/src/androidTest/java/mozilla/lockbox/RoutePresenterTest.kt
jhugman: I think by now `master` already has this in `navigator`.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,111 @@
...
+            .subscribe {
+                this.backdateNextLockTime()
+            }
+            .addTo(compositeDisposable)
+    }
+
+    override fun injectContext(context: Context) {
+        preferences = PreferenceManager.getDefaultSharedPreferences(context)
+
+        currentBootID = FileReader().readContents(Constant.App.bootIDPath)
```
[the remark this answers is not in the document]
sashei: >.<

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,111 @@
...
+    private fun updateNextLockTime(autoLockTime: Setting.AutoLockTime) {
+        storeAutoLockTimerDate(SystemClock.elapsedRealtime() + autoLockTime.ms)
+    }
+
+    private fun lockCurrentlyRequired(): Boolean {
+        return if (onNewBoot()) {
+            true
+        } else {
+            autoLockTimePassed()
+        }
```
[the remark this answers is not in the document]
sashei: :bath

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
sashei: left the diff in for logging >.<

app/src/main/java/mozilla/lockbox/support/FileReader.kt
sashei: mmm, mostly because I wanted to avoid testing our filesystem interactions inside of the autolockstore...

sashei: as far as testing goes: yes, we can mock the new instances with PowerMock.

sashei:
One passing thought that I've had: this does not take into account anything w.r.t. the login flow, so while login once has been working fine for me, I don't know what would happen with starting the login flow & backgrounding / foregrounding a few times before completion...

linuxwolf:
> One passing thought that I've had: this does not take into account anything w.r.t. the login flow, so while login once has been working fine for me, I don't know what would happen with starting the login flow & backgrounding / foregrounding a few times before completion...

Hrm, good call out.  It is (probably) likely the user will have one background/foreground cycle (FxA first-time confirmation email/code).  Worth testing for, but not sure it's worth blocking this PR on (yet)?

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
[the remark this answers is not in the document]
sashei: _fixed_

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -33,64 +38,80 @@ class AutoLockStore(
...
+    }
 
-        currentBootID = FileReader().readContents(Constant.App.bootIDPath)
-
-        preferences
-            .edit()
-            .putString(Constant.Key.bootID, currentBootID)
-            .apply()
+    private fun backdateNextLockTime() {
+        storeAutoLockTimerDate(lockingSupport.systemTimeElapsed - 1)
```
linuxwolf: What are the benefits to backdating?  Naively it seems like setting it to`0` would have been sufficient.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -51,47 +51,56 @@ class RoutePresenter(
...
     }
 
-    private fun dataStoreToRouteActions(storageState: State): Optional<RouteAction> {
+    private fun dataStoreToRouteActions(storageState: State): RouteAction {
         return when (storageState) {
             is State.Unlocked -> RouteAction.ItemList
             is State.Locked -> RouteAction.LockScreen
             is State.Unprepared -> RouteAction.Welcome
             else -> RouteAction.LockScreen
-        }.asOptional()
```
sashei: I think we might need to add this back in when @eliserichards starts sending more `State.Errored` actions (not necessarily appropriate to lock when we're offline IMO)

devinreams:
Build 1786 and 1787 on this branch are crashing on start for me on a Pixel with API 27:

```
2018-12-18 15:52:43.831 885-2118/? I/ActivityManager: Killing 22621:mozilla.lockbox/u0a158 (adj 902): crash
```

And 1785 (master) is running OK for me still.

linuxwolf:
I see the following error in an API 26 simulator image:

```
W/System.err: io.reactivex.exceptions.OnErrorNotImplementedException: The exception was not handled due to missing onError handler in the subscribe() method call. Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | /proc/sys/kernel/random/boot_id (Permission denied)
        at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:704)
        at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:701)
        at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77)
        at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:67)
        at io.reactivex.internal.operators.observable.ObservableTake$TakeObserver.onNext(ObservableTake.java:64)
        at io.reactivex.subjects.PublishSubject$PublishDisposable.onNext(PublishSubject.java:308)
        at io.reactivex.subjects.PublishSubject.onNext(PublishSubject.java:228)
        at io.reactivex.internal.operators.observable.ObservableMap$MapObserver.onNext(ObservableMap.java:62)
        at io.reactivex.internal.operators.observable.ObservableFilter$FilterObserver.onNext(ObservableFilter.java:52)
        at io.reactivex.internal.operators.observable.ObservableMap$MapObserver.onNext(ObservableMap.java:62)
W/System.err:     at io.reactivex.internal.operators.observable.ObservableFilter$FilterObserver.onNext(ObservableFilter.java:52)
        at io.reactivex.subjects.PublishSubject$PublishDisposable.onNext(PublishSubject.java:308)
        at io.reactivex.subjects.PublishSubject.onNext(PublishSubject.java:228)
        at mozilla.lockbox.flux.Dispatcher.dispatch(Dispatcher.kt:22)
        at mozilla.lockbox.presenter.ApplicationPresenter.onResume(ApplicationPresenter.kt:26)
        at android.arch.lifecycle.FullLifecycleObserverAdapter.onStateChanged(FullLifecycleObserverAdapter.java:37)
        at android.arch.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:354)
        at android.arch.lifecycle.LifecycleRegistry.forwardPass(LifecycleRegistry.java:292)
        at android.arch.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:332)
        at android.arch.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:137)
        at android.arch.lifecycle.LifecycleRegistry.handleLifecycleEvent(LifecycleRegistry.java:123)
        at android.arch.lifecycle.ProcessLifecycleOwner.activityResumed(ProcessLifecycleOwner.java:116)
        at android.arch.lifecycle.ProcessLifecycleOwner$2.onResume(ProcessLifecycleOwner.java:84)
        at android.arch.lifecycle.ReportFragment.dispatchResume(ReportFragment.java:66)
        at android.arch.lifecycle.ReportFragment.onResume(ReportFragment.java:87)
        at android.app.Fragment.performResume(Fragment.java:2657)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1319)
        at android.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1549)
        at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1611)
        at android.app.FragmentManagerImpl.dispatchMoveToState(FragmentManager.java:3039)
        at android.app.FragmentManagerImpl.dispatchResume(FragmentManager.java:3001)
        at android.app.FragmentController.dispatchResume(FragmentController.java:200)
        at android.app.Activity.performResume(Activity.java:7100)
        at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3620)
        at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3685)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2898)
        at android.app.ActivityThread.-wrap11(Unknown Source:0)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1593)
        at android.os.Handler.dispatchMessage(Handler.java:105)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6541)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
W/System.err: Caused by: java.io.FileNotFoundException: /proc/sys/kernel/random/boot_id (Permission denied)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:200)
        at java.io.FileInputStream.<init>(FileInputStream.java:150)
        at java.io.FileInputStream.<init>(FileInputStream.java:103)
        at java.io.FileReader.<init>(FileReader.java:58)
        at mozilla.lockbox.support.SimpleFileReader.readContents(SimpleFileReader.kt:9)
        at mozilla.lockbox.support.SystemLockingSupport.getCurrentBootId(LockingSupport.kt:14)
        at mozilla.lockbox.store.AutoLockStore.getCurrentBootID(AutoLockStore.kt:43)
        at mozilla.lockbox.store.AutoLockStore.storeCurrentBootId(AutoLockStore.kt:114)
        at mozilla.lockbox.store.AutoLockStore.access$storeCurrentBootId(AutoLockStore.kt:28)
        at mozilla.lockbox.store.AutoLockStore$8.accept(AutoLockStore.kt:71)
        at mozilla.lockbox.store.AutoLockStore$8.accept(AutoLockStore.kt:28)
        at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:63)
    	... 40 more
```

app/src/main/java/mozilla/lockbox/store/SettingStore.kt
```diff
@@ -89,27 +91,32 @@ open class SettingStore(
...
         unlockWithFingerprintPendingAuth = rxPrefs.getBoolean(Keys.UNLOCK_WITH_FINGERPRINT_PENDING_AUTH).asObservable()
 
-        autoLockTime = rxPrefs
+        rxPrefs
             .getString(Keys.AUTO_LOCK_TIME, Constant.SettingDefault.autoLockTime.name)
             .asObservable()
             .map {
                 Setting.AutoLockTime.valueOf(it)
             }
+            .subscribe(autoLockTime as Subject)
```
jhugman: I like this new ways of doing subscriptions.

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
```diff
@@ -14,7 +14,7 @@ import mozilla.lockbox.presenter.RoutePresenter
 
 @ExperimentalCoroutinesApi
 class RootActivity : AppCompatActivity() {
-    private var presenter: RoutePresenter = RoutePresenter(this)
+    internal var presenter: RoutePresenter = RoutePresenter(this)
```
jhugman: See above re using standard `onCreate` and `onDestroy` methods in the `simulateAppStartup` method.

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private fun lockCurrentlyRequired(): Boolean {
+        return onNewBoot() || autoLockTimeElapsed()
+    }
+
+    private fun onNewBoot(): Boolean {
+        val stored = preferences.getString(Constant.Key.bootID, null) ?: return true
+
+        return !stored.startsWith(currentBootID)
```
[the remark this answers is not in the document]
sashei: fixed in later push

app/src/androidTest/java/mozilla/lockbox/AutoLockTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        Dispatcher.shared.dispatch(LifecycleAction.Foreground)
+
+        navigator.checkAtFxALogin()
+    }
+
+    private fun simulateAppStartup() {
+        activityRule.activity.presenter.compositeDisposable.clear()
+        activityRule.activity.presenter.onViewReady()
+    }
```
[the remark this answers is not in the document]
sashei: oo I like `onDestroy` much better here

app/src/androidTest/java/mozilla/lockbox/AutoLockTest.kt
```diff
@@ -50,70 +42,71 @@ open class AutoLockTest {
...
-        simulateAppStartup()
-        navigator.gotoItemList()
-        Dispatcher.shared.dispatch(LifecycleAction.Background)
-
-        testLockingSupport.advance(Setting.AutoLockTime.FiveMinutes.ms + 1000)
-
-        Dispatcher.shared.dispatch(LifecycleAction.Foreground)
-
-        navigator.checkAtLockScreen()
+//        navigator.resetApp()
```
linuxwolf: 😞  but we'll revisit

devinreams:
I think I get a crash on Pixel, API 27 when I try to re-open the app immediately after backgrounding it, before the autolock time (eg: in less than 1 minute). 

**@sashei Since there is a workaround (open the app again) and it doesn't cause any insecure behavior (it locks), do we want want to consider addressing this in a separate issue?**

Steps:

- open app
- set autolock timer setting to 1 minute
- return the list view
- tap 'home' button to background
- tap app to re-open app
- crash! ("Reopen app again" shown in dialog)

Expected: opens to unlocked list view
Actual: shows crash dialog and re-open shows locked screen

```
2018-12-19 10:41:27.747 656-29671/? D/sound_trigger_hw: cpe_stop:[2] Exit, status=0
2018-12-19 10:41:27.747 656-29671/? D/sound_trigger_hw: sound_trigger_set_device: disable device (1) = listen-cpe-handset-mic
2018-12-19 10:41:27.754 656-29671/? D/sound_trigger_hw: cpe_dereg_sm_params:[2] Enter
2018-12-19 10:41:27.755 656-29671/? D/sound_trigger_hw: stdev_stop_recognition:[2] Exit status 0
2018-12-19 10:41:27.757 9895-9895/? I/GsaVoiceInteractionSrv: Stop Hotword Recognition status: true
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err: io.reactivex.exceptions.OnErrorNotImplementedException: The exception was not handled due to missing onError handler in the subscribe() method call. Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | navView.menuHeader must not be null
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:704)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.functions.Functions$OnErrorMissingConsumer.accept(Functions.java:701)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:67)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.drainNormal(ObservableObserveOn.java:201)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.run(ObservableObserveOn.java:255)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:119)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at android.os.Handler.handleCallback(Handler.java:790)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:99)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at android.os.Looper.loop(Looper.java:164)
2018-12-19 10:41:27.838 6412-6412/mozilla.lockbox W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:6494)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err: Caused by: java.lang.IllegalStateException: navView.menuHeader must not be null
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at mozilla.lockbox.view.ItemListFragment.updateAccountProfile(ItemListFragment.kt:179)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at mozilla.lockbox.presenter.ItemListPresenter$onViewReady$14.invoke(ItemListPresenter.kt:128)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at mozilla.lockbox.presenter.ItemListPresenter$onViewReady$14.invoke(ItemListPresenter.kt:46)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at mozilla.lockbox.presenter.ItemListPresenter$sam$io_reactivex_functions_Consumer$0.accept(Unknown Source:21)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err:     at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:63)
2018-12-19 10:41:27.839 6412-6412/mozilla.lockbox W/System.err: 	... 10 more
```

sashei:
@devinreams you are a total bug magnet, I'm not reproducing that one locally either on API 24 or 28 !!! Android 27 :shakes_fist:

Issue #117, opened 2018-10-02
Record Item telemetry events


linuxwolf:
Record telemetry events for:

* display item list
* display item details
* reveal password
* copy username to clipboard
* copy password to clipboard

Blocks #14

Issue #14, opened 2018-07-12
Add telemetry events for relevant actions in app


sandysage:
As the product owner, I want to know how users engage with the app to help inform decisions for future effort, while making sure users have control over whether or not their usage data is submitted.

## Acceptance criteria
* All events should send Telemetry event data defined by the Mozilla product, data science, and legal teams
* The user should be able to disable usage data collection (Telemetry) via a setting

Engineering email thread about testing Telemetry: https://groups.google.com/a/mozilla.com/d/msgid/mobile-all/CABupmeDgMfjz%3Dqi4m%2B23BP2jL5%2ByQV-eruCcPRUMErQLyUZRZw%40mail.gmail.com?utm_medium=email&utm_source=footer

Requires #15

Issue #205, opened 2018-11-13
Implement a `PasswordReveal` action


sashei:
Currently, in the `ItemDetailPresenter`, we don't dispatch an action (or listen for a `Store`-based source of truth) when hiding / showing the password. According to our [metrics document](https://github.com/mozilla-lockbox/lockbox-android/blob/master/docs/metrics.md), we should be tracking this hide / show behavior, which requires this implementation to use an `Action`-based flow rather than keeping everything local to the Presenter/View pair (which breaks Flux anyway, regardless of our metrics requirements!!)

Issue #121, opened 2018-10-02
Record Account telemetry events


linuxwolf:
Depends on #22 
Depends on #11 
Depends on #9 

Record telemetry events for:

* [x] The user begins the login process
* [x] The Account view is opened
* [x] ~Click "Get Started"~ -> `RouteAction.Login`
* [ ] Click login before the FxA flow begins -- this telemetry is gathered on the FxA side, not in our app
* [x] ~The app is reset (`settings_reset`) / Disconnect~ -> `object Reset : SettingAction(TelemetryEventMethod.setting_changed, TelemetryEventObject.settings_reset, null)`

Blocks #14

irrationalagent:
So we already record an event when the screen prompting to login is shown https://github.com/mozilla-lockbox/lockbox-android/blob/a911c4e0c97abe45e5455dd9a5b52853d0a37a5b/app/src/main/java/mozilla/lockbox/action/RouteAction.kt#L22

and when a user opens account settings
https://github.com/mozilla-lockbox/lockbox-android/blob/master/app/src/main/java/mozilla/lockbox/action/RouteAction.kt#L24

the only other event that I think would be plausible to add on our end is when the user actually taps the login button immediately before the flow is handed over to FxA.

eliserichards:
It looks like "Get Started" and "reset" are already in the app. I confirmed with @sashei that the 4th item ("Click login before the FxA flow begins") would happen on the FxA side and not directly from our app. Given those things, I think this issue can be closed.

PR #259, opened 2018-12-11
add telemetry events for views and copy actions

irrationalagent:
This implements a grab bag of telemetry events for RouteAction and ClipboardAction. Can do more but wanted some feedback on whether this is the right approach (questions at the end)

Fixes https://github.com/mozilla-lockbox/lockbox-android/issues/14 and fixes https://github.com/mozilla-lockbox/lockbox-android/issues/117 except for the reveal password bit which [needs an action still](https://github.com/mozilla-lockbox/lockbox-android/issues/205)

Connected to https://github.com/mozilla-lockbox/lockbox-android/issues/120, connected to https://github.com/mozilla-lockbox/lockbox-android/issues/121

## Testing and Review Notes

I made changes to https://github.com/mozilla-lockbox/lockbox-android/blob/e89a898b8ee3a1eaac79060e542bd2ac8acc2e24/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt#L20 in order to print the events to the console to verify that they are working, like:

```kotlin
open fun createEvent(category: String = "action"): TelemetryEvent {
    val evt = TelemetryEvent.create(
            category,
            eventMethod.name,
            eventObject.name,
            value
    )
    extras?.forEach { ex -> evt.extra(ex.key, ex.value.toString()) }
    println("the telemetry event method is $eventMethod") // I add this to test
    println("the telemetry event object is $eventObject") // I add this to test
    return evt
}
```

## To Do

- [ ] add unit tests
  I added some meager tests as copypasta from previous examples but there may be more to do there, guidance appreciated
- [ ] route actions decision = make them all telemetry means leave this as is
- [ ] some cleanup needed

 Also, although I've verified that this way of doing it works (thanks Matt) I'm not sure its the best or only way. For example, I don't know how to make the `EventMethod` and `EventObject` parameters optional (e.g. in `RouteAction.kt`) so as to not record events for the properties we don't really care about, e.g. we don't really need an event for the back button on https://github.com/mozilla-lockbox/lockbox-android/blob/e89a898b8ee3a1eaac79060e542bd2ac8acc2e24/app/src/main/java/mozilla/lockbox/action/RouteAction.kt#L19

### DIFF
--- a/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
@@ -9,29 +9,32 @@ package mozilla.lockbox.action
 import android.support.annotation.StringRes
 import mozilla.lockbox.flux.Action
 
-sealed class RouteAction : Action {
-    object Welcome : RouteAction()
-    object ItemList : RouteAction()
-    object Login : RouteAction()
-    object SettingList : RouteAction()
-    object AccountSetting : RouteAction()
-    object AutoLockSetting : RouteAction()
-    object Back : RouteAction()
-    object LockScreen : RouteAction()
-    object Filter : RouteAction()
-    data class ItemDetail(val id: String) : RouteAction()
-    data class OpenWebsite(val url: String) : RouteAction()
-    data class SystemSetting(val setting: SettingIntent) : RouteAction()
+sealed class RouteAction(
+    override val eventMethod: TelemetryEventMethod,
+    override val eventObject: TelemetryEventObject
+) : TelemetryAction {
+    object Welcome : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.login_welcome)
+    object ItemList : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.entry_list)
+    object Login : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.login_fxa)
+    object SettingList : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.settings_list)
+    object AccountSetting : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.settings_account)
+    object AutoLockSetting : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.settings_autolock)
+    object Back : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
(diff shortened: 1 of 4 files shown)


irrationalagent:
did my force push break the bitrise thing?

sashei:
@irrationalagent not your force push specifically.... just broken :)

app/src/test/java/mozilla/lockbox/store/TelemetryStoreTest.kt
```diff
@@ -79,6 +81,17 @@ class TelemetryStoreTest : DisposingTest() {
         eventsObserver.assertLastValueMatches {
             it.toJSON() == action.createEvent().toJSON()
         }
+        action = RouteAction.ItemList
```
linuxwolf: These additions really need to be after line 95, or it could cause failures if we add other ping upload triggers.

linuxwolf: Adding the `ClipboardAction` checks (both "username" and "password") seems worthwhile, to verify the username or password value is NOT leaked into the event.  There's some walking of the `toJSON()` involved, but that seems straight forward to do.

Otherwise, I'm not sure there's value to the other added.

app/src/main/java/mozilla/lockbox/action/RouteAction.kt
```diff
@@ -9,29 +9,32 @@ package mozilla.lockbox.action
...
-    object SettingList : RouteAction()
-    object AccountSetting : RouteAction()
-    object AutoLockSetting : RouteAction()
-    object Back : RouteAction()
-    object LockScreen : RouteAction()
-    object Filter : RouteAction()
-    data class ItemDetail(val id: String) : RouteAction()
-    data class OpenWebsite(val url: String) : RouteAction()
-    data class SystemSetting(val setting: SettingIntent) : RouteAction()
+sealed class RouteAction(
```
linuxwolf: _Discussed with @irrationalagent IRL ..._

Technically, it doesn't really matter if all `RouteAction`s are also `TelemetryAction`s.  It matters more for data science, and the resident data scientist doesn't see harm in this.

I'm inclined to leave this as changed here; @sashei @jhugman @eliserichards @ioana-farcas thoughts?

PR #273, opened 2018-12-13
initial login race condition fix

### LINKED ISSUE #244: Sync is not working on initial sign in
## Steps to reproduce

1. Open Lockbox on any device
2. Tap on "Get Started"
3. Login with a valid username and password.
4. Wait for sync

### Expected behaviour
Sync is performed and a list of logins should populate the "All entries" screen

### Actual behaviour
No sync is performed

### Device and build information
- Nexus 5 (Android 6.0.1)
- HTC Desire 820 (Android 6.0.1)
- Nokia 6 (Android 7.1.1)
- OnePlus 5T (Android 8.1.0)
- Google Pixel (Android 9)
- [Build #1316](https://app.bitrise.io/build/67245d12e783e79d)

### Notes

- Nothing was synced even after 20 minutes of waiting
- [Logcat](https://github.com/mozilla-lockbox/lockbox-android/files/2624537/log.txt)
- screen

sashei:
Fixes #244 

## To Do

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
@@ -12,7 +12,6 @@ import io.reactivex.functions.Consumer
 import io.reactivex.rxkotlin.addTo
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import mozilla.lockbox.action.AccountAction
-import mozilla.lockbox.action.DataStoreAction
 import mozilla.lockbox.action.LifecycleAction
 import mozilla.lockbox.flux.Dispatcher
 import mozilla.lockbox.flux.Presenter
@@ -36,9 +35,6 @@ class FxALoginPresenter(
             url?.let {
                 if (url.startsWith(Constant.FxA.redirectUri)) {
                     dispatcher.dispatch(AccountAction.OauthRedirect(url))
-                    // TODO: remove following line when centralized routing is implemented via:
-                    // https://github.com/mozilla-lockbox/lockbox-android/issues/144
-                    dispatcher.dispatch(DataStoreAction.Unlock)
                 }
             }
         }
--- a/app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
@@ -17,6 +17,7 @@ import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
 import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.addTo
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import mozilla.lockbox.R
 import mozilla.lockbox.action.DataStoreAction
 import mozilla.lockbox.action.RouteAction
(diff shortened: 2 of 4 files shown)


sashei:
note: we still see an ugly fxa redirect screen before going to the item list with this solution. haven't figured out a fix for that yet.

app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
```diff
@@ -36,9 +35,6 @@ class FxALoginPresenter(
             url?.let {
                 if (url.startsWith(Constant.FxA.redirectUri)) {
                     dispatcher.dispatch(AccountAction.OauthRedirect(url))
-                    // TODO: remove following line when centralized routing is implemented via:
-                    // https://github.com/mozilla-lockbox/lockbox-android/issues/144
-                    dispatcher.dispatch(DataStoreAction.Unlock)
```
jhugman: 😱

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -190,18 +192,17 @@ open class DataStore(
...
-            sync()
-        }
-
         if (backend.isLocked()) {
+            stateSubject.accept(State.Unlocking)
             backend.unlock(support.encryptionKey)
                 .asSingle(Dispatchers.Default)
-                .subscribe { _, _ -> initialSync() }
+                .subscribe { _, _ ->
+                    stateSubject.accept(State.Unlocked)
```
jhugman: Putting this here makes much more sense.

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -190,18 +192,17 @@ open class DataStore(
             return
         }
 
-        fun initialSync() {
-            unlock()
-            sync()
-        }
-
         if (backend.isLocked()) {
+            stateSubject.accept(State.Unlocking)
```
jhugman: I think we should get rid of the `.Unlocking` state.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -79,7 +81,7 @@ class RoutePresenter(
 
     private fun dataStoreToRouteActions(storageState: State): Optional<RouteAction> {
         return when (storageState) {
-            is State.Unlocking -> null
+            is State.Unlocking -> RouteAction.ItemList
```
jhugman: I'd like to kill off `.Unlocking`.

app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
linuxwolf: I did indeed LOL

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
sashei: _done_

devinreams:
>I wanted to test this with an acct with lots of logins (which you are presumably doing?)

Yes, both firefoxlockbox@gmail.com and dreams@mozilla.com

>I can file a follow-up bug, or just amend the first onboarding issue to note that we should look for it there. Thoughts?

I agree that tackling it there in that issue is fine. Do you mind amending/adding the note to that issue so its meaningful and sparks our memory? Thank you very much! 🙇

sashei:
@devinreams @linuxwolf _done_. although I realized that depending on the user's hardware / device version, it's possible that any of the three onboarding screens will be the first one they see 😆 so I've amended all of them.

PR #283, opened 2018-12-17
wait for credentials to set datastorestate

### LINKED ISSUE #271: Add Sync call during "unlock"
Related to #257

### LINKED ISSUE #275: DataStore always defaults to `Unprepared` on app launch
The routing is being thrown off because of the `BehaviorRelay` in the DataStore. It defaults to being `Unprepared` when initialized and waits for updated credentials before `Unlock`ing. Instead of assuming that the `DataStore` is always `Unprepared` on startup, we should do the state initialization based on database state or the existence of a set of credentials.

sashei:
Fixes #275 
Fixes #271 

Connected to #240 

## To Do

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
@@ -15,7 +15,6 @@ import androidx.navigation.NavController
 import androidx.navigation.Navigation
 import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
-import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.addTo
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import mozilla.lockbox.R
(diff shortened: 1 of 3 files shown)


app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -61,7 +61,7 @@ open class DataStore(
             .subscribe { state ->
                 when (state) {
                     is State.Locked -> clearList()
-                    is State.Unlocked -> updateList()
+                    is State.Unlocked -> sync()
```
jhugman: I'm nervous about this syncing on unlock.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -89,14 +88,8 @@ class RoutePresenter(
...
-    private fun accountToDataStoreActions(it: Pair<Optional<SyncCredentials>, State>): Optional<DataStoreAction> {
-        val (opt, state) = it
-
-        val credentials = opt.value ?: return DataStoreAction.Reset.asOptional()
-
-        if (state != State.Unprepared) {
-            return Optional(null)
-        }
+    private fun accountToDataStoreActions(optCredentials: Optional<SyncCredentials>): Optional<DataStoreAction> {
+        val credentials = optCredentials.value ?: return DataStoreAction.Reset.asOptional()
```
jhugman: We get a `null` `SyncCredentials` when we log out don't we?

Could you stick a comment in here just to say that a) on first run we get `null` credentials and b) on log out?

sashei:
@devinreams I didn't see any issues with my quick smoke tests locally (things like logging in, getting syncs on unlock, app staying logged in after force-quitting), but it might be good to have another pair of eyes !!

Issue #144, opened 2018-10-11
On launch select the lock screen, the entry list or the welcome screen


jhugman:
When I return to Lockbox, I want to know how to unlock my app if the prompt doesn't automatically appear.

## Acceptance criteria
* Should provide clear call to action to unlock the app when the app is locked

### Visual Design

**InVision**
Lock Screen: https://mozilla.invisionapp.com/share/3SN35UHWJDX
Unlock (Fingerprint): https://mozilla.invisionapp.com/share/3SN35UHWJDX#/310606417_Lock_Screen_-_Fingerprint

**Zeplin**
Lock Screen: https://zpl.io/25qK4En
Unlock (Fingerprint): https://zpl.io/bJq3DEJ

As a follow up to PR #133; this story shows the user the correct screen on launch, depending on the state of the app. 

 * [ ] If the app is not logged in, show the welcome screen.
 * [ ] If the app is logged in, but unlocked, show the entry list.
 * [ ] If the app is logged in, but locked, show the lock screen.

Related to #54
depends on #22

PR #285, opened 2018-12-17
Cleanup of routing work.

### LINKED ISSUE #255: Fix tech debt from routing PR #144
To follow up to #249 #144 so as to not block progress.

Scoped to just address the review comments from that pull request.

jhugman:
Fixes #255.

This PR addresses all the issues raised in the crash landed PR #144.

It does not address any bugs identified since then.


devinreams:
Thanks for addressing this all so quickly, @jhugman. 🙇

PR #286, opened 2018-12-17
185: Network connectivity warning and error handling

### LINKED ISSUE #185: Fix crash and provide useful message when  not connected to the internet
We need to provide messages to the user when content doesn't load due to connectivity issues. Along with providing messaging to the user, we also need to allow for the action of "retrying" once their connection is restored.

- fix the crash
- display dialog to the user

### Visual Design
No Internet - FxA: https://zpl.io/aRE9WQE
No Internet - Webview: https://zpl.io/VD4oqYO
No Internet - Entry List (Unpopulated): https://zpl.io/25woG3e
No Internet - Entry List (Populated): https://zpl.io/aRE9WAv

eliserichards:
Fixes #185

## Testing and Review Notes
**Note**: there are a number of followup stories to implement thorough retry logic, etc.

### review todo
- [x] change naming in `NetworkStore` etc
- [x] get rid of duplicate code in the `NetworkErrorHelper` 
- [x] check what happens on release build with `FxALoginFragment`
- [x] file followup issue for retry button logic
-- https://github.com/mozilla-lockbox/lockbox-android/issues/373
- [x] Talk about this point with James: "Thirdly, you could just put the snackbar over the view, not doing the harder job of fitting it in between the squashed view and either the toolbar or the top of the screen."
### engineering
~~- [x] add tests for NetworkStore~~ 
~~- [x] change tests for views to add new observers~~ 
~~- [x] finish visible/invisible error~~ 
~~- [x] implement sync/ ~~retest network~~ on "Retry"~~ 
~~-- when the network is down on one page and you navigate to another, Android "network unvailable" page shows up in addition to the message??~~ 
~~- [x] figure out what's going on with FxAlogin fragment~~ 
~~- [x] file issue in A-C for network availability and applicable followups for our board~~ 
-- https://github.com/mozilla-lockbox/lockbox-android/issues/355
-- https://github.com/mozilla-lockbox/lockbox-android/issues/356
-- https://github.com/mozilla-lockbox/lockbox-android/issues/369
### reviews


app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
```diff
@@ -0,0 +1,68 @@
...
+    private fun checkConnectivity() {
+        if (networkConnectivityState){
+
+            stateSubject.onNext(true)
+        }
+        else {
+            stateSubject.onNext(false)
+        }
+    }
+}
```
jhugman: I'm not sure I like this approach, here. 

We're only checking the network when you send a `CheckConnectivity` action, and then immediately getting a response back; I don't think it's very much different from simply calling the method synchronously and directly.

I _think_ there is functionality in the `ConnectivityManager` to add a listener to listen to changes in the network connection. The listener then feeds the `stateSubject`.

e.g. [`registerDefaultNetworkCallback`][1], with a fallback to a BroadcastReceiver for the right API levels.

[1]: https://developer.android.com/reference/android/net/ConnectivityManager.html#registerDefaultNetworkCallback(android.net.ConnectivityManager.NetworkCallback)

Once you get a working network connection, there are few interesting things to do here:

1. Throttle the `networkAvailability` observable. Spotty network connections cause a lot of changes of network connectivity. Would be great if we knew that the network has been up for at least a few seconds.
2. Check the liveness of the connection. Pinging a well known URL when you have a connection to check that you can actually seen the internet (rather than just the proximal network link) before you update the `networkAvailability` to `true`.

This store would be a good candidate to turn into an Android Component. Could you also file a story for that, please?

app/src/main/java/mozilla/lockbox/view/FxALoginFragment.kt
```diff
@@ -28,6 +29,18 @@ import mozilla.lockbox.support.isDebug
 
 class FxALoginFragment : BackableFragment(), FxALoginView {
     override var webViewRedirect: ((url: Uri?) -> Boolean) = { _ -> false }
+//    private var errorHelper = NetworkErrorHelper()
+//
+//    override val networkErrorVisibility: Consumer<in Boolean>
```
jhugman: The pattern we've used elsewhere is to have these methods be called on the fragment, rather than using a `Consumer`.

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
eliserichards: Todo

app/src/main/java/mozilla/lockbox/view/FxALoginFragment.kt
eliserichards: Right now, the presenters kick off `dispatcher.dispatch(NetworkAction.CheckConnectivity)` and the `Fragments` are looking at the `NetworkStore` to see what the current state is (consuming this value from the network store)

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
eliserichards: https://github.com/mozilla-lockbox/lockbox-android/issues/355

app/src/main/java/mozilla/lockbox/presenter/AppWebPagePresenter.kt
```diff
@@ -6,22 +6,40 @@
...
+    private val networkStore: NetworkStore = NetworkStore.shared,
     private val dispatcher: Dispatcher = Dispatcher.shared
 ) : Presenter() {
 
     override fun onViewReady() {
+        dispatcher.dispatch(NetworkAction.CheckConnectivity)
+
+        networkStore.networkAvailable
+            .subscribe {
+                view.handleNetworkError(it)
```
jhugman: Nit: 

```kotlin
.subscribe(view::handleNetworkError)
```

app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
```diff
@@ -50,7 +56,18 @@ class FxALoginPresenter(
...
-                dispatcher.dispatch(LifecycleAction.UseTestData)
-            }?.addTo(compositeDisposable)
+            dispatcher.dispatch(LifecycleAction.UseTestData)
+        }?.addTo(compositeDisposable)
+
+        dispatcher.dispatch(NetworkAction.CheckConnectivity)
+
+        networkStore.networkAvailable
+            .subscribe {
+                view.handleNetworkError(it)
```
jhugman: Nit: 

```kotlin
.subscribe(view::handleNetworkError)
```

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -87,12 +92,23 @@ class ItemDetailPresenter(
             .map { it.toDetailViewModel() }
             .subscribe(view::updateItem)
             .addTo(compositeDisposable)
+
+        dispatcher.dispatch(NetworkAction.CheckConnectivity)
+
+        networkStore.networkAvailable
+            .subscribe {
+                view.handleNetworkError(it)
```
jhugman: Nit: 

```kotlin
.subscribe(view::handleNetworkError)
```

app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
```diff
@@ -142,6 +147,18 @@ class ItemListPresenter(
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(view::updateAccountProfile)
             .addTo(compositeDisposable)
+
+        dispatcher.dispatch(NetworkAction.CheckConnectivity)
+
+        networkStore.networkAvailable
+            .subscribe {
+                view.handleNetworkError(it)
```
jhugman: Nit: 

```kotlin
.subscribe(view::handleNetworkError)
```

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    open lateinit var connectivityManager: ConnectivityManager
+    internal val compositeDisposable = CompositeDisposable()
+
+    private val stateSubject: ReplaySubject<Boolean> = ReplaySubject.createWithSize(1)
+    open val networkAvailable: Observable<Boolean> get() = stateSubject
+
+    // MAKE ME PRIVATE
+    private val networkConnectivityState: Boolean
+        get() = connectivityManager.activeNetworkInfo?.isConnectedOrConnecting == true
```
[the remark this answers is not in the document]
jhugman: The above naming needs a bit of attention. They are all related (`stateSubject`, `networkAvailable`, `networkConnectivityState`) but it's not clear from their names how.

e.g. `isConnected: Observable<Boolean>`,  `isConnectedSubject` and `isConnectedSync: Boolean`; it's clear they're all the same thing.

Additionally, it isn't clear why their `open`. The public API — the `Observable<Boolean>` is the only thing that is necessarily open.

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
jhugman: There are two parts to the above comment, only the second is captured by #355.

We should strive to get a nicer API, and build a sub-optimal implementation so that when we build a better implementation there is less code to change and test.

I think we should make a call to check network connectivity on `doOnSubscribe()`,  so that we don't have to manually call `dispatcher.dispatch(CheckConnectivity)`.

app/src/main/java/mozilla/lockbox/view/AppWebPageFragment.kt
```diff
@@ -62,4 +64,15 @@ class AppWebPageFragment : BackableFragment(), WebPageView {
...
+    override fun handleNetworkError(networkErrorVisibility: Boolean) {
+        if (!networkErrorVisibility) {
+            errorHelper.showWebPageNetworkError(view!!)
+        } else {
+            errorHelper.hideWebPageNetworkError(view!!)
+        }
+    }
+
+    override val retryNetworkConnectionClicks: Observable<Unit>
+        get() = view!!.networkWarning.retryButton.clicks()
```
jhugman: What does this button do?

As I understand it, it will hide the error message snack bar, but not do anything related to what we were going to use the network for.

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+        val webparams = view.webView.layoutParams as ConstraintLayout.LayoutParams
+        val toolbarMargin = view.resources.getInteger(R.integer.actionBarSize)
+        webparams.setMargins(0, toolbarMargin, 0, 0)
+        view.webView.layoutParams = webparams
+    }
+    // END FxALoginFragment
+
+    // START Item*Fragment
+    open fun showItemNetworkError(view: View) {
+        log.info("ELISE SHOW ERROR")
```
jhugman: Y'all.

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+        val toolbarPlusWarningMargin = view.resources.getInteger(R.integer.network_error_with_toolbar)
+        webParams.setMargins(0, toolbarPlusWarningMargin, 0, 0)
+        view.webView.layoutParams = webParams
+
+        view.warningMessage.text = view.resources.getString(R.string.no_internet_connection)
+    }
+
+    open fun hideLoginNetworkError(view: View) {
+        log.info("Network successfully reconnected.")
+        view.networkWarning.visibility = View.VISIBLE
```
jhugman: Should the be `VISIBLE` or `GONE`?

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+        log.info(view.resources.getString(R.string.network_connection_success))
+        view.networkWarning.visibility = View.GONE
+        view.networkWarning.layoutParams.height = R.dimen.hiddenNetworkError
+
+        // set margin of entriesView to show only toolbar
+        val marginLayoutParams =  ViewGroup.MarginLayoutParams(view.refreshContainer.entriesView.layoutParams)
+        marginLayoutParams.leftMargin = 0
+        marginLayoutParams.topMargin = 0
+        marginLayoutParams.rightMargin = 0
+        marginLayoutParams.bottomMargin = 0
```
jhugman: Is `setMargins(0,0,0,0)` available here?

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+        log.info("ELISE SHOW ERROR")
+        log.error(view.resources.getString(R.string.networkWarningMessage))
+        view.networkWarning.visibility = View.VISIBLE
+        view.warningMessage.text = view.resources.getString(R.string.no_internet_connection)
+    }
+
+    open fun hideItemListNetworkError(view: View) {
+        log.info(view.resources.getString(R.string.network_connection_success))
+        view.networkWarning.visibility = View.GONE
+        view.networkWarning.layoutParams.height = R.dimen.hiddenNetworkError
```
jhugman: I don't think this line does anything, since you are setting it to `GONE` in the line above.

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
```diff
@@ -0,0 +1,106 @@
...
+        log.info("Network successfully reconnected.")
+        view.networkWarning.visibility = View.VISIBLE
+
+        // set margin of webview to show only toolbar
+        val webparams = view.webView.layoutParams as ConstraintLayout.LayoutParams
+        webparams.setMargins(0, 0, 0, 0)
+        view.webView.layoutParams = webparams
+    }
+    // END AppWebPageFragment
+}
```
jhugman: There's lots of duplicated code in this file. Can we reduce this down, perhaps by making the XML more uniform. e.g. make `ConstraintLayout` the default layout we use everywhere.

You can also pass in parameters to the show and hide methods e.g. 

```
open fun hideNetworkError(networkWarning: View, squashedView: View) {
   networkWarning.visibility = View.GONE
   val layoutParams = squashedView.layoutParams
   if (layoutParams is ConstraintLayout.LayoutParams) {
        layoutParams.setMargins(0,0,0,0)
   }
   squashedView.layoutParams = layoutParams
}
```

Thirdly, you could just put the snackbar over the view, not doing the harder job of fitting it in between the squashed view and either the toolbar or the top of the screen.

app/src/main/res/layout/fragment_fxa_login.xml
```diff
@@ -14,21 +14,35 @@
...
 
     <WebView
             android:id="@+id/webView"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:layout_marginTop="?attr/actionBarSize"
-            android:layout_weight="1"/>
+            app:layout_constraintVertical_chainStyle="packed"
+            app:layout_constraintTop_toBottomOf="@id/toolbar"
+            app:layout_constraintBottom_toTopOf="@id/skipFxA"
```
jhugman: This could be very risky here, when the `skipFxA` button is not visible. e.g. in release builds.

app/src/main/res/layout/fragment_item_detail.xml
```diff
@@ -13,14 +13,20 @@
...
+             tools:text="@string/no_internet_connection"/>
 
     <android.support.v7.widget.CardView
             xmlns:card_view="http://schemas.android.com/apk/res-auto"
             android:id="@+id/card_view"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_gravity="center"
-            card_view:layout_constraintTop_toBottomOf="@id/toolbar">
+            card_view:layout_constraintTop_toBottomOf="@id/networkWarning">
```
jhugman: What happens when `networkWarning` disappears?

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
eliserichards: yes, miss on my part

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
eliserichards: `View.GONE` isn't actually making it entirely disappear (don't ask me why, I couldn't figure it out) :(

app/src/main/java/mozilla/lockbox/view/NetworkErrorHelper.kt
eliserichards: I worked for a while on making the layouts more uniform, but found that it would take a lot of time to reformat every fragment that way, so I abandoned that in favor of this not-so-nice formatting.

I'd like to talk about the 3rd point with you (I don't think I'm sure about how that would fit into the designs).

app/src/main/res/layout/fragment_fxa_login.xml
eliserichards: good point. I'll change that around

app/src/main/res/layout/fragment_item_detail.xml
eliserichards: It constrains to the "height=0" network warning

eliserichards:
@jhugman : 

> 1. The NetworkStore api should be less reliant on `CheckConnectivity` actions, and more reliant on `doOnSubscribe` and `doOnDispose` of the `isConnected` observable.
✅ 

> 2. I don't know what the retry button is supposed to do. This is presenter dependent, and should be more than displaying or hiding the error message, but I don't know what.
Right now the retry button is simply checking the connection status again. The thought being that, if the network is not connected, the user can recheck the connection once they reconnect to wifi or what have you. This should be addressed in a followup issue, making the retry logic more explicit (e.g. re-sync, reload a web page, etc). Also see @linuxwolf 's input on retry logic: https://github.com/mozilla-lockbox/lockbox-android/issues/185#issuecomment-448378427

> 3. The `NetworkErrorHelper` could be made a little smaller. How can we delete some of this code by changing the XML in someway.
I agree that the fragments _should_ be made more consistent. I addressed this in a comment [above](https://github.com/mozilla-lockbox/lockbox-android/pull/286/#discussion_r246883080), but I found that it was taking up too much time and abandoned it for now. Maybe address this in a followup issue as well? I'm not sure what the priority of doing a refactor like that would be.

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
```diff
@@ -0,0 +1,68 @@
...
+                    is NetworkAction.CheckConnectivity -> {
+                        checkConnectivity()
+                    }
+                }
+            }
+            .addTo(compositeDisposable)
+
+        stateSubject.subscribe { state ->
+                networkAvailable.doOnNext { state }
+        }.addTo(compositeDisposable)
```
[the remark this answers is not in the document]
eliserichards: removed this

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
eliserichards: Added #369 for the longer-term A-C stuff, and changed #355 to reflect that first part about adding a listener to the network connectivity manager

app/src/main/java/mozilla/lockbox/store/NetworkStore.kt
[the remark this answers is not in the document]
eliserichards: re: `open` -- leftovers from testing, going to remove that

re: naming -- I was getting confused with all of the "state" in names. I like your suggestion about making the names about the connection

eliserichards:
@jhugman, question about `doOnSubscribe()`...

Currently, in my `Presenters`, I have the following:
`networkStore.isConnected
            .doOnSubscribe { dispatcher.dispatch(NetworkAction.CheckConnectivity) }
            .subscribe(view::handleNetworkError)
            .addTo(compositeDisposable)`

We discussed this morning that `.doOnSubscribe{ ... }` should be moved to the `NetworkStore` instead of being called in every Presenter. I am stuck with where/how to implement this. My own theory makes me think that we should have something in the NetworkStore, where the`init` accesses the Observable `isConnected` like this:
`init {
        dispatcher.register
             ... .addTo(compositeDisposable)

        isConnected
            .doOnSubscribe { dispatcher.dispatch(NetworkAction.CheckConnectivity) }
            .addTo(compositeDisposable)
    }`

Is this the right path to go down? Or am I going down a rabbit hole and missing the point of the `doOnSubscribe`?

jhugman:
> @jhugman, question about `doOnSubscribe()`...
> 
> Currently, in my `Presenters`, I have the following:
> 
> ```kotlin
> networkStore.isConnected
>              .doOnSubscribe { dispatcher.dispatch(NetworkAction.CheckConnectivity) }
>              .subscribe(view::handleNetworkError).addTo(compositeDisposable)
> ```
> 
>  My own theory makes me think that we should have something in the NetworkStore, where the`init` accesses the Observable `isConnected` like this:
> 
> ```kotlin
> init {             
>              ... 
>              isConnected
>                      .doOnSubscribe { dispatcher.dispatch(NetworkAction.CheckConnectivity) }
>                      .addTo(compositeDisposable)
> }
> ```
> 
> Is this the right path to go down? Or am I going down a rabbit hole and missing the point of the `doOnSubscribe`?

Yes, this looks like the right way to go about it.

The `doOnSubscribe` method is registering a closure to be run each time a new subscription is made on `isConnected`. We're putting it in the store so you can subscribe in the presenters, and immediately have it come back with an up-to-date state; without the presenters having to worry how that all happens.

If you think: presenters are subscribing to the store and the store is looking after the subscriptions; then the store will want to do something each time it receives a subscription.

--

Another thing you may consider is eliminating the `dispatcher.dispatch` step, and just call the `checkConectivity()` method directly. Other reviewers may have other opinions about whether to miss out the dispatch step or not.

eliserichards:
@nickbrandt there is a followup issue for the crash you're seeing. That was the "temporary" fix that we implemented before we released. The issue https://github.com/mozilla-lockbox/lockbox-android/issues/356 is On Deck right now, hopefully to be fixed soon. My PR covers once you get to the welcome screen and then turn the network off/on. Could you try it again and let me know if you're still getting crashes?

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
```diff
@@ -58,21 +72,36 @@ class FxALoginPresenterTest : DisposingTest() {
...
     lateinit var subject: FxALoginPresenter
 
     @Before
     fun setUp() {
+        whenCalled(networkStore.isConnected).thenReturn(isConnected)
         whenCalled(accountStore.loginURL).thenReturn(loginURLSubject)
 
         PowerMockito.whenNew(AccountStore::class.java).withAnyArguments().thenReturn(accountStore)
-        Dispatcher.shared.register.subscribe(dispatcherObserver)
+        dispatcher.register.subscribe(dispatcherObserver)
```
jhugman: Yes!

jhugman: Yes!

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
```diff
@@ -116,7 +145,15 @@ class FxALoginPresenterTest : DisposingTest() {
         if (isDebug()) {
             (view.skipFxAClicks as PublishSubject).onNext(Unit)
 
-            dispatcherObserver.assertValue(LifecycleAction.UseTestData)
+            dispatcherObserver.assertValueAt(0, LifecycleAction.UseTestData)
```
jhugman: Ok.

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
```diff
@@ -58,21 +72,36 @@ class FxALoginPresenterTest : DisposingTest() {
...
     fun setUp() {
+        whenCalled(networkStore.isConnected).thenReturn(isConnected)
         whenCalled(accountStore.loginURL).thenReturn(loginURLSubject)
 
         PowerMockito.whenNew(AccountStore::class.java).withAnyArguments().thenReturn(accountStore)
-        Dispatcher.shared.register.subscribe(dispatcherObserver)
+        dispatcher.register.subscribe(dispatcherObserver)
 
-        subject = FxALoginPresenter(view, accountStore = accountStore)
+        networkStore.connectivityManager = connectivityManager
```
jhugman: Not sure I understand this. The `networkStore` is a mock, and you're giving it mock `connectivityManager`?

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
```diff
@@ -116,7 +145,15 @@ class FxALoginPresenterTest : DisposingTest() {
...
             (view.skipFxAClicks as PublishSubject).onNext(Unit)
 
-            dispatcherObserver.assertValue(LifecycleAction.UseTestData)
+            dispatcherObserver.assertValueAt(0, LifecycleAction.UseTestData)
         }
     }
+
+    @Test
+    fun `network error visibility is correctly being set`() {
+        val value = view.networkAvailable
```
jhugman: Nit: you can inline this.

app/src/test/java/mozilla/lockbox/presenter/TestApplication.kt
```diff
@@ -0,0 +1,15 @@
...
+
+@ExperimentalCoroutinesApi
+class TestApplication : LockboxApplication() {
+
+    init {
+        log.info("Using TestApplication")
+    }
+
+    override val unitTesting = true
+}
```
jhugman: I'm not sure this is useful, given that we have `isTesting` methods.

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -74,6 +77,7 @@ class LockboxApplication : Application() {
             FxASyncDataStoreSupport.shared,
             ClipboardStore.shared,
             FingerprintStore.shared,
+            NetworkStore.shared,
```
jhugman: :+1:

nickbrandt:
@eliserichards I tried downloading the build again this morning, and I'm still experiencing crashes on going to the FxA screen with my network available.

linuxwolf:
I just tried the release binary, from scratch (uninstall Lockbox, install from bitrise, launch).  It crashes when I try "Getting Started".  It appears, for some reason, that `ConstraintLayout` is expected but missing:

```
    --------- beginning of crash
2019-01-22 11:18:43.653 19065-19065/mozilla.lockbox E/AndroidRuntime: FATAL EXCEPTION: main
    Process: mozilla.lockbox, PID: 19065
    android.view.InflateException: Binary XML file line #20: Binary XML file line #7: Error inflating class android.support.constraint.ConstraintLayout
    Caused by: android.view.InflateException: Binary XML file line #7: Error inflating class android.support.constraint.ConstraintLayout
    Caused by: java.lang.ClassNotFoundException: Didn't find class "android.support.constraint.ConstraintLayout" on path: DexPathList[[zip file "/system/framework/org.apache.http.legacy.boot.jar", zip file "/data/app/mozilla.lockbox--1x_wLv-dOodeaFwv1R4sw==/base.apk"],nativeLibraryDirectories=[/data/app/mozilla.lockbox--1x_wLv-dOodeaFwv1R4sw==/lib/arm64, /data/app/mozilla.lockbox--1x_wLv-dOodeaFwv1R4sw==/base.apk!/lib/arm64-v8a, /system/lib64]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:134)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
        at android.view.LayoutInflater.createView(LayoutInflater.java:606)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:790)
        at android.view.LayoutInflater.parseInclude(LayoutInflater.java:965)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:859)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:824)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:515)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:423)
        at mozilla.lockbox.view.FxALoginFragment.onCreateView(FxALoginFragment.kt:45)
        at androidx.fragment.app.Fragment.performCreateView(Fragment.java:2440)
        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:885)
        at androidx.fragment.app.FragmentManagerImpl.addAddedFragments(FragmentManagerImpl.java:2078)
        at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1852)
        at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1808)
        at androidx.fragment.app.FragmentManagerImpl.execPendingActions(FragmentManagerImpl.java:1709)
        at androidx.fragment.app.FragmentManagerImpl$1.run(FragmentManagerImpl.java:147)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
eliserichards: you can't create an instance of a ConnectivityManager, it has to be mocked. And the network store's connectivity manager has to be instantiated or it will break

app/src/test/java/mozilla/lockbox/presenter/TestApplication.kt
eliserichards: The application was failing on Leak Canary, even before it reached the isTesting method. I'm going to file a followup for this

PR #311, opened 2018-12-20
Improve generation of Device-local Encryption Keys

### LINKED ISSUE #228: Securely generate the encryption key for new SQLCipher databases
As part of #221, the encryption key needs to be generated. 

Currently, [there are two naive ways of doing this implemented][1].

Minimally, one of these ways needs to be selected, and audited.

Maximally, this should be re-implemented.

Generating this key would be a good candidate to go into DataProtect. 

/cc @linuxwolf 

[1]: https://github.com/mozilla-lockbox/lockbox-android/blob/6057ba8a4c58f3d59c81b2d605a3d34fe75ce173/app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt#L44

linuxwolf:
Fixes #228

Improves how the SQLCipher key is generated to provide the entropy we desire (>= 256 bits).

## Testing and Review Notes

This requires a review of the implementation; it's not directly verifiable per se.

* [x] review from one of (@rfk / @vladikoff / @eoger)
* [x] review from one of (@ncalexan  / @thomcc)

## To Do

- ~add “WIP” to the PR title if pushing up but not complete nor ready for review~
- [ ] add unit tests
- ~request the "UX" team perform a design review (if/when applicable)~
- ~check on the [accessibility](https://mozilla-lockbox.github.io/lockbox-android/accessibility/) of any added UI~

### DIFF
--- a/app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
+++ b/app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
@@ -32,26 +32,23 @@ class FxASyncDataStoreSupport(
         }
 
         // val encryptionKey = randomUUID().toString()
-        val encryptionKey = generateRandomString(50)
+        val encryptionKey = generateRandomString(32)
         preferences.putString(Constant.Key.encryptionKey, encryptionKey)
 
         encryptionKey
     }
 
-    /*
-     * This serves as a placeholder, and should definitely not be considered secure.
-     * TODO make this cryptographically interesting; add method to KeyStore/DataProtect
-     */
     private fun generateRandomString(keyLength: Int): String {
-        val charPool = "0123456789ABCDEF".toCharArray()
-
-        val bytes = ByteArray(keyLength)
+        val bytes = ByteArray(keyLength * 2)
         val random = SecureRandom()
         random.nextBytes(bytes)
 
         return bytes
             .map {
-                charPool[(it.toInt() and 0xFF % charPool.size)]
+                val full = it.toInt()
+                val hi = (full and 0xf0) ushr 4
+                val lo = full and 0x0f
+                "%h%h".format(hi, lo)
             }
             .joinToString("")
     }
--- a/app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
+++ b/app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
@@ -14,6 +14,24 @@ import mozilla.components.service.sync.logins.AsyncLoginsStorageAdapter
 import mozilla.lockbox.store.ContextStore
 import java.security.SecureRandom
 
+/**
+ * @param keyStrength The strength ofthe generated key in bits
+ */
+internal fun generateEncryptionKey(keyStrength: Int): String {
+    val bytes = ByteArray(keyStrength / 8)
+    val random = SecureRandom()
+    random.nextBytes(bytes)
+
+    return bytes
+        .map {
+            val full = it.toInt()
+            val hi = (full and 0xf0) ushr 4
+            val lo = full and 0x0f
+            "%x%x".format(hi, lo)
+        }
+        .joinToString("")
+}
+
 class FxASyncDataStoreSupport(
     private val preferences: SecurePreferences = SecurePreferences.shared
 ) : DataStoreSupport, ContextStore {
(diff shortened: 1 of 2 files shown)


linuxwolf:
@rfk / @vladikoff / @eoger : would one of you check to make sure I'm not doing something egregiously dumb?

@ncalexan / @thomcc : ditto?

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
```diff
@@ -32,26 +32,23 @@ class FxASyncDataStoreSupport(
...
+        val bytes = ByteArray(keyLength * 2)
         val random = SecureRandom()
         random.nextBytes(bytes)
 
         return bytes
             .map {
-                charPool[(it.toInt() and 0xFF % charPool.size)]
+                val full = it.toInt()
+                val hi = (full and 0xf0) ushr 4
+                val lo = full and 0x0f
```
rfk: It seems weird to me that you have to manually split out the nibbles like this, but the logic appears sound.

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
```diff
@@ -32,26 +32,23 @@ class FxASyncDataStoreSupport(
...
         preferences.putString(Constant.Key.encryptionKey, encryptionKey)
 
         encryptionKey
     }
 
-    /*
-     * This serves as a placeholder, and should definitely not be considered secure.
-     * TODO make this cryptographically interesting; add method to KeyStore/DataProtect
-     */
     private fun generateRandomString(keyLength: Int): String {
```
rfk: IIUC, the returned string will contain this many *bytes of entropy*, but will not be this length.  Is this a potential API footgun where callers might expect the returned string to be of the specified length?

linuxwolf: your understanding is correct, I'll change the argument name

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
linuxwolf: from my searching this seemed to be the best available short of adding another dependency 🙃

linuxwolf:
Thanks @rfk !

>  Perhaps we can make this go away entirely by moving the "make the key safe for use in sqlcipher" step into the logins component?

That could help, yes.  It would be ideal to provide a `ByteArray` directly rather than encoding it somehow 😄 

> I'd welcome a bug on our side if that makes sense to you.

Will do, for the benefit of the future if nothing else.

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
thomcc: Splitting the nibbles out is how you do this almost always. It's what the dependency would almost certainly do as well.

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
```diff
@@ -32,26 +32,26 @@ class FxASyncDataStoreSupport(
...
         val random = SecureRandom()
         random.nextBytes(bytes)
 
         return bytes
             .map {
-                charPool[(it.toInt() and 0xFF % charPool.size)]
+                val full = it.toInt()
+                val hi = (full and 0xf0) ushr 4
+                val lo = full and 0x0f
+                "%h%h".format(hi, lo)
```
thomcc: I just was checking to see if this formatted upper case or lower case (for https://github.com/mozilla/application-services/issues/499) and apparently %h isn't the hex formatter, it's the hashcode formatter. This is, of course, further argument that 499 is important, but it's worth noting that it won't be compatible in that way.

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
```diff
@@ -32,26 +32,23 @@ class FxASyncDataStoreSupport(
         }
 
         // val encryptionKey = randomUUID().toString()
```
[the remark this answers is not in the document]
linuxwolf: no, we don't; will remove.

app/src/main/java/mozilla/lockbox/support/FxASyncDataStoreSupport.kt
linuxwolf: 🤦‍♂️  thanks for the catch.

PR #337, opened 2018-12-27
[WIP] upgrade to a-c 0.34.0, accountstore tests

### LINKED ISSUE #241: Upgrade Android-Components to 0.35.1
Update Android Components to version 0.35.1 that provides needed improvements to the `FirefoxAccount` class.

**Breaking Changes**
Access tokens can only be obtained for a single scope, rather than all scopes.
* Lockbox **MUST** only ask for `Constant.FxA.oldSyncScope` when obtaining the access token
* Lockbox **MUST** still ask for all scopes when authorizing via `getOAuthURL` + webView.

## Original Description

After https://github.com/mozilla/application-services/issues/252, Lockbox for Android should be able to unit test the a-s logins Rust code (at least on Linux and macOS/Darwin).

This ticket tracks wiring that up on the Lockbox side.

sashei:
Fixes #241

_(**Required**: this reference (one or many) will be closed upon merge. Ideally it has the acceptance criteria and designs for features or fixes related to the work in this Pull Request.)_

Connected to #???

_(Optional: other issues or pull requests related to this, but merging should not close it)_

## Testing and Review Notes

## Screenshots or Videos

## To Do

- [ ] add unit tests


sashei:
😭 evidently we do _not_ have the cross-compilation working for running the`fxa-client` library on MacOS or Linux yet, so the account store tests are still on hold for now. communication whoops

sashei:
This only tracks the upgrade to `0.34.0` now.

build.gradle
```diff
@@ -9,7 +9,7 @@
 buildscript {
     ext.android_support_version = '28.0.0'
     ext.kotlin_version = '1.3.0'
-    ext.android_components_version = '0.33.0'
+    ext.android_components_version = '0.34.1'
```
linuxwolf: From other discussions/issues, I think we really want to go to `0.35.1` or higher (latest as of 2019-01-02 should be `0.36`)

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -106,66 +105,47 @@ open class AccountStore(
...
-                .delay(1L, TimeUnit.SECONDS)
-                .subscribe({
-                    val syncCredentials = generateSyncCredentials(it, isNew)
-                    syncCredentialSubject.onNext(syncCredentials.asOptional())
-                }, {
-                    dispatcher.dispatch(RouteAction.Dialog.NoNetworkDisclaimer)
-                })
-                .addTo(compositeDisposable)
-        }
+        fxa.getAccessToken(Constant.FxA.oldSyncScope)
```
linuxwolf: Strictly speaking, obtaining a fresh access token is something that will happen with some regularity (whenever LoginsStorage "throws" a `SyncAuthInvalidException` error ... which should exist in `0.34` or higher).  Is it worth breaking this call out into its own method that can/will be wired up to some sort of "refresh token" action?

build.gradle
linuxwolf: I do understand that.  Looking through their commits, `0.34.1` had all the breaking changes that affect us.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
linuxwolf: I know I asked the question a leading way, but it is really a question.  To phrase it differently, should refreshing the token also refresh profile information?

I can see arguments for both sides.  The more I think about it, I slightly prefer refreshing both (if we're at this state, we've already pushed the radio to its most active state anyways, so might as well get as much data as we can).

sashei: I'm curious what the intended behavior for this is on the client side, maybe @rfk or @vladikoff have thoughts?

rfk: @linuxwolf's suggestion to refresh both sync token and profile info seems fine to me.  Longer-term we could try to get you a push notification whenever the profile info changes, if that seems worthwhile, since we're going to need to build that out for the reference-browser anyways.

PR #349, opened 2019-01-07
Ensure that the lock screen is maintained when the app is supposed to be locked.

### LINKED ISSUE #260: When in locked state app can re-open to list view
During demo by @jhugman we observed the app being manually "Locked" and then backgrounded.

Upon re-opening, it opened the entry list view (not the locked/unlock button view).

jhugman:
There are serveral parts to this, each of which is slow and difficult. 

 - [x] Ensure that the AutoLockTests work.
 - [x] Write failing new tests for the Locking edge cases
 - [x] Correct the application code such that the new tests and `AutoLockTests` all work.

This PR: 
 * updates the use of the android support library to androidx.
 * stabilizes a lot of tests by moving to the AndroidX Test Coordinator. 
 * re-writes `AutoLockStore` to broadcast the need for locking, rather than autolocking would fire here if it needed here. This makes auto locking a little more encapsulated.
 * `AutoLockStoreTest` needed changing in order to make that work.
 * It also moves a lot of logic out of the `RoutePresenter` to the various stores.
    - `DataStore`
    - `RouteStore`
    - `AutoLockStore`

Fixes #260

### DIFF
--- a/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
@@ -19,6 +19,7 @@ import io.reactivex.subjects.ReplaySubject
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import mozilla.lockbox.action.DataStoreAction
 import mozilla.lockbox.action.LifecycleAction
+import mozilla.lockbox.action.RouteAction
 import mozilla.lockbox.flux.Dispatcher
 import mozilla.lockbox.robots.accountSettingScreen
 import mozilla.lockbox.robots.disconnectDisclaimer
@@ -31,8 +32,10 @@ import mozilla.lockbox.robots.securityDisclaimer
 import mozilla.lockbox.robots.settings
 import mozilla.lockbox.robots.welcome
 import mozilla.lockbox.store.DataStore
+import mozilla.lockbox.store.RouteStore
 import mozilla.lockbox.view.RootActivity
 import org.junit.Assert
+import java.lang.Thread.sleep
 
 @ExperimentalCoroutinesApi
 class Navigator {
@@ -62,6 +65,10 @@ class Navigator {
         Dispatcher.shared.dispatch(LifecycleAction.UserReset)
         blockUntil(DataStore.State.Unprepared)
         checkAtWelcome()
+    }
+
+    fun blockUntil(vararg states: RouteAction) {
+        blockUntil(RouteStore.shared.routes, *states)
     }
 
     fun blockUntil(vararg states: DataStore.State) {
@@ -95,7 +102,8 @@ class Navigator {
             fxaLogin { tapPlaceholderLogin() }
         } else {
             Dispatcher.shared.dispatch(LifecycleAction.UseTestData)
-            blockUntil(DataStore.State.Unlocked)
+            log.info("blocking for the routes")
+            blockUntil(RouteStore.shared.routes, RouteAction.ItemList)
         }
         checkAtItemList()
     }
--- a/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
@@ -14,6 +14,8 @@ import android.support.test.espresso.intent.Intents
 import android.support.test.rule.ActivityTestRule
 import br.com.concretesolutions.kappuccino.custom.intent.IntentMatcherInteractions.sentIntent
 import br.com.concretesolutions.kappuccino.custom.intent.IntentMatcherInteractions.stubIntent
+import io.reactivex.Observable
+import io.reactivex.subjects.ReplaySubject
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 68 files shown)


app/src/main/java/mozilla/lockbox/view/DialogFragment.kt
```diff
@@ -7,12 +7,12 @@
...
 import android.os.Bundle
-import android.support.annotation.StringRes
-import android.support.v4.app.DialogFragment
+import androidx.annotation.StringRes
+import androidx.fragment.app.DialogFragment
 import android.view.View
 import mozilla.lockbox.flux.Presenter
 
-open class DialogFragment : DialogFragment() {
+open class DialogFragment : androidx.fragment.app.DialogFragment() {
```
sashei: nit: import this rather than having the long class name

app/src/main/java/mozilla/lockbox/view/FilterFragment.kt
```diff
@@ -38,10 +38,10 @@ class FilterFragment : BackableFragment(), FilterView {
         presenter = FilterPresenter(this)
         val view = inflater.inflate(R.layout.fragment_filter, container, false)
 
-        val layoutManager = LinearLayoutManager(context)
+        val layoutManager = androidx.recyclerview.widget.LinearLayoutManager(context)
```
sashei: nit: import over long class name

app/src/main/res/layout/fragment_fingerprint_dialog.xml
```diff
@@ -4,15 +4,15 @@
   ~ file, You can obtain one at http://mozilla.org/MPL/2.0/.
   -->
 
-<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
-    android:paddingTop="32dp">
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
sashei: 👾

app/src/main/res/layout/fragment_item_list.xml
```diff
@@ -4,21 +4,21 @@
...
   -->
-<android.support.v4.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/appDrawer"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    tools:context=".view.ItemListFragment">
+<androidx.drawerlayout.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
+                                           xmlns:app="http://schemas.android.com/apk/res-auto"
```
sashei: 👾

app/src/androidTest/java/mozilla/lockbox/Navigator.kt
```diff
@@ -38,14 +42,46 @@ class Navigator {
...
+                Dispatcher.shared.dispatch(DataStoreAction.Unlock)
+                blockUntil(DataStore.State.Unlocked)
+            }
+
+            is DataStore.State.Unprepared -> {
+                return
+            }
+
+            else -> {
+                // NOP
```
eliserichards: Should this be empty?

app/src/androidTest/java/mozilla/lockbox/RoutePresenterTest.kt
```diff
@@ -30,11 +29,6 @@ open class RoutePresenterTest {
     @Rule @JvmField
     val activityRule: ActivityTestRule<RootActivity> = ActivityTestRule(RootActivity::class.java)
 
-    @Before
-    fun setUp() {
-        navigator.resetApp(activityRule)
```
eliserichards: Why do we no longer have to use `resetApp`?

eliserichards:
LGTM! 

I see some tests failing on the branch locally (`SettingListAdapterTest`, `SectionedAdapterTest`), but I assume those are the reason that bitrise is red.

app/src/androidTest/java/mozilla/lockbox/RoutePresenterTest.kt
jhugman: Moving over to the test coordinator means we're now clearing the data in the app before each test: in essence doing a better job of resetting the app that we were doing here.

Result: more stable tests!

app/src/main/java/mozilla/lockbox/view/DialogFragment.kt
jhugman: :+1:

app/src/androidTest/java/mozilla/lockbox/Navigator.kt
jhugman: Yes. The `// NOP` is short for noop. We should probably make this comment more like '// do nothing`, minimally.

Issue #382, opened 2019-01-17
Display launch icon for Web Address field on Entry Detail screen


nickbrandt:
For consistency with action icons on the Username and Password fields (and parity with what we have now on iOS), we need to add the launch/open icon to the Web Address field on the Entry Detail screen.

PR #400, opened 2019-01-23
Make entire username and password views clickable 

### LINKED ISSUE #281: Set the entire UN/PW field to be tappable (for Copy)
Currently, the Copy icons are the only way in which someone can copy a given username or password.

Given the low level associations of icons to a corresponding action, we would like to have the entire row (sans the show/hide password eye) invoke the copy action. The only risk (which was mitigated by #49) would have been inadvertent copying, but this will allow people who don't associate the copy icon with the action a more 'frictionless' way to tap to copy. 

screen shot 2018-12-17 at 11 09 02 am

devinreams:
Fixes #281
Depends on #399

To aim to be atomic, and since these changes are in _slightly_ different territory, this adds on to the commit at #382 #399.

## Testing and Review Notes

1. open item detail view
2. tap to confirm the entire username/password row is 'clickable'
3. (accessibility) confirm the voice over instructions explain the row will "copy your username/password"

Note: because the text area appears on screen before the "reveal password" button, it may be confusing, but the accessibility / tab order is intentional:

1. web address (open in browser)
2. username (copy)
3. password (copy)
4. password (reveal)

## Screenshots or Videos

android_emulator_-_pixel_2_api_28_5554

## To Do

- [ ] add unit tests
- [x] fix the "password is always revealed" logic bug I introduced
- [ ] put the "null input" in a better way/right place? if applicable...


devinreams:
Thanks for the speedy reviews @nickbrandt. I definitely introduced a bug or two here and will look to fix it soon...

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -117,9 +118,10 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
 var EditText.readOnly: Boolean
     get() = this.isFocusable
     set(readOnly) {
-        this.isFocusable = !readOnly
+        this.isFocusable = true
         this.isFocusableInTouchMode = !readOnly
-        this.isClickable = !readOnly
+        this.inputType = InputType.TYPE_NULL
+        this.isClickable = true
```
jhugman: Now you're making specializations specific to the detail screen, you should probably put these into their own method. Still use this property, but forcing values for `inputType` `isFocusable` and `isClickable` should go after the call to `readOnly`.

app/src/main/res/values/strings.xml
```diff
@@ -83,6 +83,8 @@
     <!-- This message appears to link to a web page explaining how to make changes to the entry details -->
     <string name="learn_to_edit_entry">Learn how to edit this entry</string>
 
+    <!-- This describes an icon to screenreaders that the button tapped will open the web address in a web browser -->
+    <string name="launch_hostname_content_description">Open web address in web browser</string>
```
jhugman: 👍

devinreams:
>When I navigate to the Entry Detail, the password is already revealed (as seen in screenshot)

Ah yes, I don't know how to fix this but can tell it's a side effect of my changing to the edit text (instead of the button) as the observable so I need... a different default state or explictly setting the hidden state somewhere?

**I would appreciate help pairing or pointers from @mozilla-lockbox/mobile-engineering here.** ✋ 

>as well as the Web Address is underlined in blue and has a green label. The tap targets work as intended, just experiencing the password and styling issues.

Whoops, yeah that was me being too aggressive making changes.

**This should be fixed now.** ✅

devinreams:
Thanks for the patch @linuxwolf 03304cf

@eliserichards when you have a moment, how do these changes look to you? can you try building/testing it yourself too and see if its working as expected?

I'll do the same and ask for help from UX to test again too...

And we'll need to get the lint warnings fixed, that may mean simply updating the dependencies or just waiting a few moments since they're about to get bumped at #413 anyway...

linuxwolf:
Waiting for #413 would be my preference; I think it's practically ready anyway.

PR #402, opened 2019-01-24
add telemetry events for settings changes

irrationalagent:
Fixes https://github.com/mozilla-lockbox/lockbox-android/issues/120

adds telemetry events for `SettingsAction` .
I had to make`Setting.AutoLockTime.seconds` non-private in order to pass the value into the telemetry event `value` field. If that's badness let me know

Also, e.g. in https://github.com/mozilla-lockbox/lockbox-android/blob/3abe5d5a6dd8217a726a3b442ddac21a34bb83f1/app/src/main/java/mozilla/lockbox/action/SettingAction.kt#L17 

how would i do something like `getString(sortOrder.titleId)` ? so I can include the human readable string in in the `value` field for the telemetry event? doing `sortOrder.titleId.toString()` just returns the integer code corresponding to `R.string.all_entries_a_z` which isn't great for telemetry

## Testing and Review Notes

I made changes to https://github.com/mozilla-lockbox/lockbox-android/blob/e89a898b8ee3a1eaac79060e542bd2ac8acc2e24/app/src/main/java/mozilla/lockbox/action/TelemetryAction.kt#L20 in order to print the events to the console to verify that they are working, like:

```kotlin
open fun createEvent(category: String = "action"): TelemetryEvent {
    val evt = TelemetryEvent.create(
            category,
            eventMethod.name,
            eventObject.name,
            value
    )
    extras?.forEach { ex -> evt.extra(ex.key, ex.value.toString()) }
    println("the telemetry event method is $eventMethod") // I add this to test
    println("the telemetry event object is $eventObject") // I add this to test
    println("the telemetry event value is $value") // I add this to test
    return evt
}
```

## Screenshots or Videos

e.g.
image

## To Do

- [X] double check the original issue to confirm it is fully satisfied
- [X] add testing notes and screenshots in PR description to help guide reviewers
- [X] add unit tests _this could be better_


app/src/main/java/mozilla/lockbox/action/SettingAction.kt
```diff
@@ -8,21 +8,25 @@ package mozilla.lockbox.action
...
+    data class ItemListSortOrder(val sortOrder: Setting.ItemListSort) : SettingAction(TelemetryEventMethod.setting_changed, TelemetryEventObject.settings_item_list_order, null)
+    data class AutoLockTime(val time: Setting.AutoLockTime) : SettingAction(TelemetryEventMethod.setting_changed, TelemetryEventObject.settings_autolock_time, time.seconds.toString())
+    object Reset : SettingAction(TelemetryEventMethod.setting_changed, TelemetryEventObject.settings_reset, null)
 }
 
 class Setting {
     enum class AutoLockTime(
         @StringRes val stringValue: Int,
-        private val seconds: Long
+        val seconds: Long
```
jhugman: This is fine, as long as this is immutable, which it is :)

PR #413, opened 2019-01-31
Authenticate via fingerprint or PIN to autofill when the app is locked

### LINKED ISSUE #216: Authenticate via fingerprint or PIN to autofill when the app is locked
### Acceptance Criteria

Attempting to autofill with a credential from Lockbox while the app is locked will prompt the `DialogFragment` with either fingerprint or PIN authentication.

#### Documentation

[User Authentication](https://developer.android.com/reference/android/service/autofill/AutofillService#user-authentication)

#### Implementation
This involves adding the [`setAuthentication`](https://developer.android.com/reference/android/service/autofill/FillResponse.Builder#setAuthentication(android.view.autofill.AutofillId[],%20android.content.IntentSender,%20android.widget.RemoteViews)) call to the `FillResponse` object when returning from the `buildFillResponse` method in the `LockboxAutofillService`.

#### Possible Gotchas
In order to get a list of potential credentials, the `DataStore` has to be unlocked, so you will need to capture its lock state before unlocking it, use _that_ state to determine when to add the authentication view in the `buildFillResponse` method, and lock it again if the authentication step fails.

depends on #214

ioana-farcas:
Fixes #216 

## Testing and Review Notes
In the current state, this works for authenticating using PIN (when app does not enter in a weird state, when there are no entries in the list). For authentication using fingerprint, the only reason it is not working (as far as I can see), is the fact that DataStore.State == null. Don't know exactly why, I will try tomorrow to take a look at this.
@jhugman I think you will take over, the code needs a little refactoring..

## Screenshots or Videos

## To Do

- [x] add unit tests


app/src/main/java/mozilla/lockbox/autofill/FillResponseBuilder.kt
```diff
@@ -0,0 +1,103 @@
...
+        val sender = AuthActivity.getAuthIntentSender(context, this)
+        val autofillIds = arrayOf(parsedStructure.usernameId, parsedStructure.passwordId)
+
+        builder.setAuthentication(autofillIds, sender, searchPresentation)
+    }
+
+    fun buildFilteredFillResponse(context: Context, passwords: List<ServerPassword>): FillResponse? {
+        val possibleValues = passwords.filter {
+            val credential = URL(it.hostname)
+            (credential.host.endsWith(hostname, true) || hostname.endsWith(credential.host, true))
```
sashei: I think some of this logic will change / need to incorporate @linuxwolf 's work at #425

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -59,15 +58,15 @@ open class LockboxApplication : Application() {
...
-        // this needs to be done after injectContext, as
-        // SyncDataStoreSupport needs to find the database
-        // path from the context
-        val support = if (isTesting()) {
-            FixedDataStoreSupport.shared
-        } else {
-            FxASyncDataStoreSupport.shared
+        // This list of stores need to be constructed
+        // in the given order. e.g. AccountStore dispatches DataStoreActions.
+        val orderedStores = listOf(
```
sashei: :-/

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -96,6 +97,18 @@ open class AccountStore(
             }
             .addTo(compositeDisposable)
 
+        // Moves credentials from the AccountStore, into the DataStore.
+        syncCredentials.map {
```
sashei: ooo

we don't typically have stores dispatch actions?
would it make more sense to have the datastore subscribe to the account store, or ?

app/src/main/java/mozilla/lockbox/store/AutoLockStore.kt
```diff
@@ -80,6 +80,12 @@ class AutoLockStore(
             // push lockrequired if required
             .map { lockCurrentlyRequired() }
             .subscribe(autoLockActivated as Subject)
+
+        autoLockActivated
+            .filter { it }
+            .map { DataStoreAction.Lock }
+            .subscribe(dispatcher::dispatch)
```
sashei: similarly, in iOS we ended up having the datastore read the appropriate lock state and take action accordingly, rather than dispatch from a store.

sashei: are we going to be able to remove the `AccountStore` -> `DataStoreAction` part of the `RootPresenter` because of this?

sashei: ...ah I see you did that

app/src/main/java/mozilla/lockbox/presenter/AuthPresenter.kt
```diff
@@ -0,0 +1,94 @@
...
+    private val responseBuilder: FillResponseBuilder,
+    private val dispatcher: Dispatcher = Dispatcher.shared,
+    private val fingerprintStore: FingerprintStore = FingerprintStore.shared,
+    private val settingStore: SettingStore = SettingStore.shared,
+    private val dataStore: DataStore = DataStore.shared,
+    private val lockedStore: LockedStore = LockedStore.shared
+) : Presenter() {
+
+    override fun onViewReady() {
+        Observables.combineLatest(settingStore.unlockWithFingerprint, dataStore.state)
```
sashei: `dataStore.state` is the source of truth, not `autoLockStore.lock`

app/src/main/java/mozilla/lockbox/presenter/AuthPresenter.kt
```diff
@@ -0,0 +1,94 @@
...
+            .filter { it }
+            .subscribe {
+                unlock()
+            }
+            .addTo(compositeDisposable)
+
+        dataStore.state
+            .filter { it == DataStore.State.Unlocked }
+            .switchMap { dataStore.list }
+            .subscribe { passwords ->
```
sashei: The hairy part IIRC is that we need to provide the authentication response _along with_ the list of passwords to the `AutofillService`... have you found a way around that?

app/src/main/java/mozilla/lockbox/autofill/FillResponseBuilder.kt
jhugman: Yis.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
jhugman: I think stores dispatching actions is fine. Subscribing to others makes me more queasy. Let's discuss.

app/src/main/java/mozilla/lockbox/presenter/AuthPresenter.kt
jhugman: We need to authenticate before unlocking & showing a list of passwords, rather than showing a list of passwords and then authenticating.

We should also note that the list is empty when the data store is listed.

app/src/main/res/values/strings.xml
```diff
@@ -231,4 +230,11 @@
     <!-- This description of the app convenience is shown on the final onboarding screen. -->
     <string name="convenience_description">Unlock the app with ease using your fingerprint</string>
 
+    <!-- This is the prompt when autofill is triggered, and Lockbox is locked. -->
+    <string name="autofill_authenticate_cta">Tap to unlock</string>
+    <!-- This is the error message when autofill is triggered but no domain is detected. This is unlikely. -->
+    <string name="autofill_error_no_hostname">Unexpected hostname format</string>
+    <!-- This is the prompt when autofill is triggered, Lockbox is unlocked, and no passwords are detected. -->
+    <string name="autofill_search_cta">Tap to search</string>
```
linuxwolf: UX designs call for this to be "Search Lockbox"

app/src/main/res/values/strings.xml
```diff
@@ -231,4 +230,11 @@
     <!-- This description of the app convenience is shown on the final onboarding screen. -->
     <string name="convenience_description">Unlock the app with ease using your fingerprint</string>
 
+    <!-- This is the prompt when autofill is triggered, and Lockbox is locked. -->
+    <string name="autofill_authenticate_cta">Tap to unlock</string>
```
linuxwolf: I suspect @mozilla-lockbox/ux would prefer "Unlock Lockbox"

sashei: UNLOCK YOUR FIRELOCK FOXBOX

Issue #16, opened 2018-07-12
Metrics dashboard


sandysage:
Requires #15 
Requires #14

sandysage:
Moving to 'in progress' based on the status update @irrationalagent gave in stand up today.

devinreams:
Dashboard was quickly shared today. Any other updates (extra telemetry events, cut-off at launch) will happen as part of regular review and maintenance. I think this is safe to close as done!

Thanks @irrationalagent

PR #418, opened 2019-02-04
update metrics.md with currently collected events

irrationalagent:
We needed to updated the metrics doc to reflect the events that are currently collected, so this is me doing that. no changes to code.

Fixes: NA

Connected to #https://github.com/mozilla-lockbox/lockbox-android/issues/16


irrationalagent:
I'll be honest, its not anything I've ever actually used for iOS. In the interest of not adding more telemetry than we need, I'm going to remove that bit from the doc for now. If there's ever a use case, I'll add it.

PR #428, opened 2019-02-07
launch biometrics from lock screen when returning from background

### LINKED ISSUE #323: Automatically prompt to unlock when returning from background
Consider automatically throwing the unlock (prompt for finger, pin) when coming back from background so user doesn't have to tap the button, like on iOS

Design/UX note: For device PIN/Pattern/Password, it would be ideal to first show the Lockbox Lock Screen when foregrounding, then slide to the Device PIN/Pattern/Password screen, as we handle it in iOS (so user's aren't confused by opening the app directly to the device auth screen)

Depends on #435

sashei:
Fixes #323 

## Testing and Review Notes
Similar to iOS, this _will not_ automatically launch the biometrics or PIN prompts when returning from the background after _hard locking_ (i.e., locking using the Lock Now button). This simplifies implementation significantly and was a product concession there; hopefully it's acceptable here as well.

Cases:
- app cold start, locked by auto lock time => launch device auth
- app foreground, locked by auto lock time from anywhere else => launch device auth
- "lock now" tapped, lock screen => NO device auth launch
- app foreground after "lock now" tapped => NO device auth launch
- app cold start after "lock now" tapped => launch device auth
- app foreground after beginning device auth flow and backgrounding / cancelling => NO device auth launch
- app cold start after beginning device auth flow and hard-quitting => device auth launch

## To Do

- [x] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
+++ b/app/src/main/java/mozilla/lockbox/LockboxApplication.kt
@@ -7,8 +7,8 @@
 package mozilla.lockbox
 
 import android.app.Application
-import androidx.lifecycle.ProcessLifecycleOwner
 import android.os.Build
+import androidx.lifecycle.ProcessLifecycleOwner
 import com.squareup.leakcanary.LeakCanary
 import io.sentry.Sentry
 import io.sentry.android.AndroidSentryClientFactory
@@ -19,7 +19,6 @@ import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.log.sink.AndroidLogSink
 import mozilla.lockbox.presenter.ApplicationPresenter
 import mozilla.lockbox.store.AccountStore
-import mozilla.lockbox.support.AutoLockSupport
 import mozilla.lockbox.store.ClipboardStore
 import mozilla.lockbox.store.ContextStore
 import mozilla.lockbox.store.DataStore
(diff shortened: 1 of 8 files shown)


sashei:
@nickbrandt I don't know that we have any control over that part of the UI; does it happen when tapping the `unlock` button as well? IMO that sounds like a separate issue / exploration from this PR if you'd like to file a ticket for it.

nickbrandt:
@sashei yes, it is the same when tapping the "Unlock" button as well.

I've also noticed another issue: When presented with the device auth, if you tap the Back button (which brings you back to the unlock screen), the "Unlock" button no longer functions. I have to hard quit out of the app to get the prompt again.

When presented with the Fingerprint dialog, close it (tap outside of modal, tap "Cancel", hit Back button), I have the same issue. Though backgrounding and then foregrounding brings up the prompt again.

sashei:
@nickbrandt the unlock button not working is definitely related to this branch; it should always work even if you cancel out of the PIN or fingerprint entry. funny, I'm not seeing it on my physical device but I did replicate that in my emulator. let me look into it!

sashei:
@nickbrandt nice catch! it had to do with the "Don't keep activities" setting on my physical device D:

I think I've fixed it for both cases (the process-limited app killing version, and the normal app behavior verison). let me know how it looks to you!

PR #439, opened 2019-02-14
Dispatch telemetry for a revealed password

### LINKED ISSUE #205: Implement a `PasswordReveal` action

linuxwolf:
Adds an action dispatched when the user would be revealing an item's password.

This patch matches the details as well as the intent.  The store is limited to just data that actual action can be taken on (password visibility), and roughly matches the implementation in iOS.

Fixes #205 

## Testing and Review Notes

Unit tests confirm the action is dispatched when revealed, and not dispatched when hidden.

**NOTE** this implementation can record a `reveal_password` telemetry event when the user shows and hides the password, ostensibly matching the behavior in iOS.

## To Do

- [x] “WIP” to the PR title if pushing up but not complete nor ready for review
- [x] add unit tests


linuxwolf:
@mozilla-lockbox/mobile-engineering this PR is ready for review at your convenience.

@irrationalagent I'd like your opinion on the telemetry recording; as far as I can tell, this duplicates the behavior in iOS, but it also records a `reveal_password` telemetry event each time the user taps the button -- whether it is in fact revealing or concealing the password.

app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
```diff
@@ -0,0 +1,52 @@
...
+    init {
+        dispatcher.register
+            .filterByType(RouteAction::class.java)
+            .subscribe { action ->
+                when (action) {
+                    is RouteAction.ItemDetail -> _passwordVisible.onNext(false)
+                    is RouteAction.ItemList -> _passwordVisible.onNext(false)
+                }
+            }
+            .addTo(compositeDisposable)
```
jhugman: AFAIUI this is "each time you go from item list to item detail or back again, make the password invisible."

Is there a simpler way we can achieve this?

app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
```diff
@@ -0,0 +1,52 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *
```
jhugman: Nit: whitespace

app/src/main/java/mozilla/lockbox/action/ItemDetailAction.kt
```diff
@@ -0,0 +1,17 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ *
```
jhugman: Nit: whitespace.

app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
linuxwolf: nothing else was obvious.  Also note that this pattern is almost exactly what iOS does today.

app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
linuxwolf: oops!
Aaaand it's in the just-merged template, too.

app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
jhugman: See my comment re: `ItemListPresenter` dispatching an `ItemDetailAction.View(id)` action, which itself dispatches the `RouteAction`.

linuxwolf:
> Looks good. Default is not visible, meaning if a client generates this event at least once they viewed the password at least once correct? That's all that's important from an analysis perspective.

Correct, it starts hidden, and the first tap makes it visible.

Issue #435, opened 2019-02-12
Implement explicit app initialization behavior


linuxwolf:
Implement an explicit initialization stage to start the app's "source of truth".  This may involve any of the following:

* An explicit SplashFragment (+ SplashPresenter) which replaces itself with the "true" starting fragment (locked, welcome, itemlist)
* Code that examines necessary stored data (e.g., Stores that read from prefs?) and takes steps to ensure all stores are fully initialized (e.g., The DataStore is Locked/Unlocked/Unprepared)
* Stores that fully initialize themselves dispatch actions that will further initialize other stores
* Stores read (e.g., `take(1).subscribe(...)`) state from other stores to initialize their state

Currently when the application starts up, the initial state of some stores is left undetermined until other actions are dispatched and handled.  This has resulted in initialization being distributed and in some cases implicit, and is believed to be the cause of many of the inconsistency and "broken-ness" experienced.

Related to #260 
Related to #323 
Related to #360 
Related to #415 

Cases:
- [x] startup with no account => welcome
- [x] foreground with no account after ~5 minutes => stays on previous screen ?
- [x] startup with no account, login, onboarding => itemlist
- [x] startup with account, later than lock time => lock screen
- [x] foreground with account, later than lock time => lock screen
- [x] manually lock, hit back button, foreground => lock screen
- [x] startup with account, before lock time => itemlist
- [x] foreground with account, before lock time => stays on previous screen
- [x] unlock with fullscreen system prompt from lock screen => itemlist
- [x] unlock with fingerprint from lock screen => itemlist

sashei:
Thoughts from 🍐 w/ @eliserichards :

Ideally, we'd like the initialization of app state to correspond with an initialization of the `DataStore`. Rather than having the `DataStore` wait passively for dispatched initialization actions, shift as much of the querying / initialization / foregrounding logic as possible into the `DataStore.init` method. The other relevant stores we need to have in this scenario are the `AccountStore` and the `AutoLockStore`. Ideas:

- `AutoLockStore`: change the `autoLockRequired` variable to a `Boolean` rather than an `Observable<Boolean>` and have the `DataStore` query that only when appropriate (i.e., foreground, app start). 
- `AccountStore`: either have this store dispatch the `UpdateCredentials` action a la #413 , OR have the `DataStore`subscribe to the `syncCredentials` observable and do state management appropriately

sashei:
[...]

PR #446, opened 2019-02-15
Make autolock and autofill play nicely.

### LINKED ISSUE #442: Autolocking Stability Part II — Autofill
Related to #435 and follwing up from #413, this issue ensures that autolock and autofill intersect. #435 fixed a lot of stability issues by moving much of auto-lock into the datastore. This issue: 

 * documents the lifecycle actions for foreground and background.
 * introduces two new lifecycle actions, for autofill service that correspond to existing foreground and background.
 * verifies that background and foreground events are no longer as important as they were, since they cannot be relied upon to be the same as the user's model of backgrounding and foregrounding the app 
 * verifies that the autolock timer is applied to successive autofill instances, without needing to go to launch the app.
 * moves AccountStore to use lifecyleStore instead of using dispatchers.

jhugman:
Fix #442. 

This builds on #435 and #413 to stabilize autolock in the context of autofill.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/LockboxAutofillService.kt
+++ b/app/src/main/java/mozilla/lockbox/LockboxAutofillService.kt
@@ -17,6 +17,7 @@ import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.addTo
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import mozilla.lockbox.action.LifecycleAction
 import mozilla.lockbox.autofill.FillResponseBuilder
 import mozilla.lockbox.autofill.ViewNodeNavigator
 import mozilla.lockbox.extensions.dump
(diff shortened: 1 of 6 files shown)


app/src/main/java/mozilla/lockbox/LockboxAutofillService.kt
```diff
@@ -40,7 +42,12 @@ class LockboxAutofillService(
     private var compositeDisposable = CompositeDisposable()
     private val pslSupport = PublicSuffixSupport.shared
 
+    override fun onConnected() {
+        dispatcher.dispatch(LifecycleAction.AutofillStart)
```
sashei: 👍 👍 👍

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -45,6 +45,7 @@ import kotlin.coroutines.CoroutineContext
 
 @ExperimentalCoroutinesApi
 open class AccountStore(
+    private val lifecycleStore: LifecycleStore = LifecycleStore.shared,
```
sashei: 🛁

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -104,15 +104,18 @@ open class DataStore(
     }
 
     private fun setupAutoLock() {
+        val foreground = arrayOf(LifecycleAction.Foreground, LifecycleAction.AutofillStart)
```
sashei: cooooool

PR #473, opened 2019-02-26
Removing functions to set username copy ability, adding back to fragm…

eliserichards:
…ent.

Fixes username copy

## Testing and Review Notes
Breaking out the `showUsernamePlaceholder()`/`showUsername()` into functions that are called from the presenter was preventing the fragment from setting the click and focus correctly. This was preventing the user from successfully copying their username. 

Removed code, replaced with a flag (`showUsernamePlaceholder`) that is being set in the presenter to be used in the `updateItem()` function: 
```
    override fun showUsernamePlaceholder() {
        view!!.btnUsernameCopy.setColorFilter(resources.getColor(R.color.white_60_percent))
        view!!.isClickable = false
        view!!.isFocusable = false
        inputUsername.setText(" ", TextView.BufferType.NORMAL)
    }

    override fun showUsername(username: String) {
        inputUsername.isClickable = true
        inputUsername.isFocusable = true
        inputUsername.setText(username, TextView.BufferType.NORMAL)
    }
```

## To Do

- [ ] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
@@ -34,11 +34,10 @@ interface ItemDetailView {
     val hostnameClicks: Observable<Unit>
     val learnMoreClicks: Observable<Unit>
     var isPasswordVisible: Boolean
-    fun updateItem(item: ItemDetailViewModel)
+    fun updateItem(item: ItemDetailViewModel, showUsernamePlaceholder: Boolean)
     fun showToastNotification(@StringRes strId: Int)
     fun handleNetworkError(networkErrorVisibility: Boolean)
     //    val retryNetworkConnectionClicks: Observable<Unit>
-    var showUsernamePlaceholder: Boolean
 }
 
 @ExperimentalCoroutinesApi
@@ -96,8 +95,8 @@ class ItemDetailPresenter(
             .doOnNext { credentials = it }
             .map { it.toDetailViewModel() }
             .subscribe {
-                view.showUsernamePlaceholder = it.username.isNullOrEmpty()
-                view.updateItem(it)
+                val showUsernamePlaceholder = it.username.isNullOrEmpty()
+                view.updateItem(it, showUsernamePlaceholder)
             }
             .addTo(compositeDisposable)
 
--- a/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
@@ -38,8 +38,7 @@ interface ItemDetailView {
     fun showToastNotification(@StringRes strId: Int)
     fun handleNetworkError(networkErrorVisibility: Boolean)
     //    val retryNetworkConnectionClicks: Observable<Unit>
-    fun showUsernamePlaceholder()
-    fun showUsername(username: String)
+    var showUsernamePlaceholder: Boolean
 }
 
 @ExperimentalCoroutinesApi
@@ -56,7 +55,7 @@ class ItemDetailPresenter(
 
     override fun onViewReady() {
         handleClicks(view.usernameCopyClicks) {
-            if (!it.username.isNullOrEmpty()) {
+            if (!it.username.isNullOrEmpty() && it.username != " ") {
                 dispatcher.dispatch(ClipboardAction.CopyUsername(it.username.toString()))
                 dispatcher.dispatch(DataStoreAction.Touch(it.id))
                 view.showToastNotification(R.string.toast_username_copied)
@@ -97,11 +96,7 @@ class ItemDetailPresenter(
             .doOnNext { credentials = it }
             .map { it.toDetailViewModel() }
             .subscribe {
-                if (it.username.isNullOrEmpty()) {
-                    view.showUsernamePlaceholder()
-                } else {
-                    view.showUsername(it.username)
-                }
+                view.showUsernamePlaceholder = it.username.isNullOrEmpty()
                 view.updateItem(it)
             }
             .addTo(compositeDisposable)
--- a/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
@@ -34,7 +34,7 @@ interface ItemDetailView {
     val hostnameClicks: Observable<Unit>
     val learnMoreClicks: Observable<Unit>
     var isPasswordVisible: Boolean
-    fun updateItem(item: ItemDetailViewModel, showUsernamePlaceholder: Boolean)
+    fun updateItem(item: ItemDetailViewModel)
     fun showToastNotification(@StringRes strId: Int)
     fun handleNetworkError(networkErrorVisibility: Boolean)
     //    val retryNetworkConnectionClicks: Observable<Unit>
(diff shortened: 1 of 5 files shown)


app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -102,6 +90,20 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
         inputLayoutPassword.isHintAnimationEnabled = false
 
         inputUsername.readOnly = true
+        when (showUsernamePlaceholder) {
+            true -> {
+                view!!.btnUsernameCopy.setColorFilter(resources.getColor(R.color.white_60_percent))
+                view!!.isClickable = false
+                view!!.isFocusable = false
+                inputUsername.setText(" ", TextView.BufferType.NORMAL)
```
jhugman: Make this a string resource.

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -102,6 +90,20 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
...
+            }
+            false -> {
+                inputUsername.isClickable = true
+                inputUsername.isFocusable = true
+                inputUsername.setText(item.username, TextView.BufferType.NORMAL)
+            }
```
jhugman: This `when` statement is difficult to read. Perhaps it could do with a comment or two. 

1. On one hand you're dealing with the view, on the other, you're dealing with the `inputUsername`.
2. `when` on booleans still look weird to me. This may be a stylistic things.

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
eliserichards: The resources aren't accessible from the ItemDetailFragment because it doesn't have any Context

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
eliserichards: I could change it to an if/else. Might be easier to read. I am not changing the original code, just grouping it based on the username being available.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -55,7 +54,7 @@ class ItemDetailPresenter(
 
     override fun onViewReady() {
         handleClicks(view.usernameCopyClicks) {
-            if (!it.username.isNullOrEmpty()) {
+            if (!it.username.isNullOrEmpty() && it.username != " ") {
```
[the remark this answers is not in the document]
eliserichards: I like that! 👏

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -96,11 +95,7 @@ class ItemDetailPresenter(
             .doOnNext { credentials = it }
             .map { it.toDetailViewModel() }
             .subscribe {
-                if (it.username.isNullOrEmpty()) {
-                    view.showUsernamePlaceholder()
-                } else {
-                    view.showUsername(it.username)
-                }
+                view.showUsernamePlaceholder = it.username.isNullOrEmpty()
```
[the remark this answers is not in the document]
eliserichards: I like this too!

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -102,6 +88,18 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
         inputLayoutPassword.isHintAnimationEnabled = false
 
         inputUsername.readOnly = true
+
+        if (showUsernamePlaceholder) {
+            view!!.btnUsernameCopy.setColorFilter(resources.getColor(R.color.white_60_percent))
+            view!!.isClickable = false
+            view!!.isFocusable = false
+            inputUsername.setText(getString(R.string.empty_space), TextView.BufferType.NORMAL)
```
jhugman: You don't need the `getString` here.

```kotlin
inputUsername.setText(R.string.empty_space, TextView.BufferType.NORMAL)
```

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -102,6 +88,18 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
...
+        } else {
+            inputUsername.isClickable = true
+            inputUsername.isFocusable = true
+            inputUsername.setText(item.username, TextView.BufferType.NORMAL)
+        }
```
jhugman: I'm still confused about this. I know that you're only copying code from where it was before to here. However, I don't understand the rationale for the code in the first place. 

```kotlin
btnUsernameCopy.setColorFilter(resources.getColor(R.color.white_60_percent))
view!!.isClickable = false
view!!.isFocusable = false
inputUsername.setText(R.string.empty_space, TextView.BufferType.NORMAL)
```

Intuitively, each side of this if/else should look very similar, if only to reverse the changes that successive calls make.

Shouldn't the `inputUsername` be non-clickable, and non-focusable, rather than the whole `view!!`?

app/src/test/java/mozilla/lockbox/presenter/ItemDetailPresenterTest.kt
```diff
@@ -184,7 +176,7 @@ class ItemDetailPresenterTest {
                 fakeCredentialNoUsername.hostname,
                 fakeCredentialNoUsername.username,
                 fakeCredentialNoUsername.password
-            )
+            ), true
```
jhugman: Nit: whitespace. Put on a separate line.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -97,12 +95,8 @@ class ItemDetailPresenter(
...
             .map { it.toDetailViewModel() }
             .subscribe {
-                if (it.username.isNullOrEmpty()) {
-                    view.showUsernamePlaceholder()
-                } else {
-                    view.showUsername(it.username)
-                }
-                view.updateItem(it)
+                val showUsernamePlaceholder = it.username.isNullOrEmpty()
+                view.updateItem(it, showUsernamePlaceholder)
```
jhugman: I wonder whether the `showUsernamePlaceholder` should really be calculated either in the view model or the view itself?

jhugman: Looking at line 57, it seems that this should be in the view model, as a calculated property/function `hasUsername`.

eliserichards: I tried putting this in the view model as well, but like the original commit, this broke the copy ability for some reason. That's why I put it back in the fragment/presenter.

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
eliserichards: Good catch. This was leftover copy/paste from when this was a separate method. should be `inputUsername`

eliserichards:
@jhugman Since this has been open for a week already and was meant as a quick fix, I would like to address any further comments you have in another PR as long as you are ok with this fixing the original issue that copy was not functioning

sashei:
@eliserichards I took the liberty of toying with this a little bit; I think the "copyable"-ness of the username was actually being affected by the order in which `inputUsername.readOnly` is called. 

For every other field, you set the `readOnly` property _before_ making them `isClickable` and `isFocusable`. To fix the bug, just move the `view.updateItem` call to before the username-specifc method calls. Alternately, move the `inputUsername.readOnly = true` to the top of each of the username methods.

This will preserve the appropriate separation of concerns between the view and the presenter.

PR #474, opened 2019-02-26
472: Adding autofill route if fingerprint onboarding is skipped

### LINKED ISSUE #472: When device security is not set, the Autofill onboarding screen is skipped on first run
If a user does not have device security set up on their phone (and they choose to skip setting it up from the dialog on Welcome screen), the Autofill onboarding screen is bypassed and they are brought directly to the "You're all set!" screen after signing in to FxA.

Expected behavior: The user would bypass the Fingerprint onboarding screen only (as there is no Fingerprint set), but would still see the Autofill onboarding screen.

eliserichards:
Fixes #472 

## To Do

- [ ] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
@@ -39,7 +39,8 @@ class FxALoginPresenter(
     private val dispatcher: Dispatcher = Dispatcher.shared,
     private val networkStore: NetworkStore = NetworkStore.shared,
     private val accountStore: AccountStore = AccountStore.shared,
-    private val fingerprintStore: FingerprintStore = FingerprintStore.shared
+    private val fingerprintStore: FingerprintStore = FingerprintStore.shared,
+    private val settingStore: SettingStore = SettingStore.shared
 ) : Presenter() {
     fun isRedirectUri(uri: String?): Boolean = uri?.startsWith(Constant.FxA.redirectUri) ?: false
 
@@ -80,7 +81,7 @@ class FxALoginPresenter(
     private fun triggerOnboarding() {
         if (fingerprintStore.isFingerprintAuthAvailable) {
             dispatcher.dispatch(RouteAction.Onboarding.FingerprintAuth)
-        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && SettingStore.shared.autofillAvailable) {
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && settingStore.autofillAvailable) {
             dispatcher.dispatch(RouteAction.Onboarding.Autofill)
         } else {
             dispatcher.dispatch(RouteAction.Onboarding.Confirmation)
--- a/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
@@ -7,6 +7,7 @@
 package mozilla.lockbox.presenter
 
 import android.net.Uri
+import android.os.Build
 import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.rxkotlin.addTo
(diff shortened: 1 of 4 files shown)


app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
```diff
@@ -77,6 +80,10 @@ class FxALoginPresenter(
     private fun triggerOnboarding() {
         if (fingerprintStore.isFingerprintAuthAvailable) {
             dispatcher.dispatch(RouteAction.Onboarding.FingerprintAuth)
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            if (SettingStore.shared.autofillAvailable) {
+                dispatcher.dispatch(RouteAction.Onboarding.Autofill)
+            }
```
jhugman: What happens in Android versions at or above `O`, but where `autofillAvailable` is `false`?

eliserichards: I believe that's what it should do. If autofill isn't available then we shouldn't tell them to use autofill 🤔

app/src/test/java/mozilla/lockbox/presenter/FxALoginPresenterTest.kt
```diff
@@ -154,7 +167,6 @@ class FxALoginPresenterTest : DisposingTest() {
         Assert.assertEquals(OnboardingStatusAction(true), dispatcherObserver.values()[0])
         val redirectAction = dispatcherObserver.values()[1] as AccountAction.OauthRedirect
         Assert.assertEquals(url.toString(), redirectAction.url)
```
sashei: can you add a test for the extra case here rather than removing a test case?

app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
```diff
@@ -77,6 +81,8 @@ class FxALoginPresenter(
     private fun triggerOnboarding() {
         if (fingerprintStore.isFingerprintAuthAvailable) {
             dispatcher.dispatch(RouteAction.Onboarding.FingerprintAuth)
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && settingStore.autofillAvailable) {
+            dispatcher.dispatch(RouteAction.Onboarding.Autofill)
```
jhugman: 🤘

app/src/main/java/mozilla/lockbox/store/SettingStore.kt
```diff
@@ -46,7 +46,7 @@ open class SettingStore(
 
     private lateinit var preferences: SharedPreferences
     @RequiresApi(Build.VERSION_CODES.O)
-    private lateinit var autofillManager: AutofillManager
+    open lateinit var autofillManager: AutofillManager
```
jhugman: 👾

This is making `AutofillManager` part of the public interface for this store. If I understand correctly, this is to make it easier to test. 

I think probably the best way of solving this problem is to mock out the autofill manager and context, then inject it using `injectContext`. The [`SettingStoreTest` shows this][1] quite nicely.

[1]: https://github.com/mozilla-lockbox/lockbox-android/blob/5c321cff1031f75f0e6be5b0b731ca3b846e776e/app/src/test/java/mozilla/lockbox/store/SettingStoreTest.kt#L195

eliserichards: Ahh I think I tried every way of mocking the AutofillManager and SettingStore except `whenCalled(context.getSystemService(AutofillManager::class.java)).thenReturn(autofillManager)`. Thanks!

app/src/main/java/mozilla/lockbox/presenter/FxALoginPresenter.kt
```diff
@@ -20,13 +21,15 @@ import mozilla.lockbox.flux.Presenter
...
 import mozilla.lockbox.store.FingerprintStore
 import mozilla.lockbox.store.NetworkStore
+import mozilla.lockbox.store.SettingStore
 import mozilla.lockbox.support.Constant
 
 interface FxALoginView {
     var webViewRedirect: ((url: Uri?) -> Boolean)
     val skipFxAClicks: Observable<Unit>?
-//    val retryNetworkConnectionClicks: Observable<Unit>
+    //    val retryNetworkConnectionClicks: Observable<Unit>
```
sashei: 👾

PR #483, opened 2019-02-28
No issue: use local jacoco configuration

mcomella:
@sashei mentioned codecov was not reporting the same results as local code coverage runs so I created this PR to use jacoco through a local configuration like we do on FFES, FFTV, Focus, etc. thinking this might affect the outcome. I wanted to test to see if it improved coverage. My assumption is that the plugin has the jacoco version pinned to an older version.

**N.B.:** when I ran this locally coverage improved about 2% over the plugin jacoco declaration.

Note: with this PR, the coverage command changed to `./gradlew -Pcoverage jacocodebugtestreport`

However, there are many other moving pieces:
- Is codecov running its own coverage metrics (afaik, it just reads the output files though)?
- Is bitrise running a different coverage command than @sashei was locally?
- Is bitrise running its own coverage commands separate from the repositories setup? (is that possible?)

Let's hope codecov runs on my PRs too. :)


mcomella:
I can't get the bitrise logs but I'm going to guess it's mad that the jacoco command changed. @sashei Is there any way to modify the jacoco command for this run?

devinreams:
>odecov was not reporting the same results as local code coverage

Acknowledging that local vs codecov (website _or_ build status here) could be explained by this, I also wanted to note: some of the difference we see at codecov.io vs the "app" build status on GitHub here would be because we re-defined it to exclude the un-tested AccountStore code:

https://github.com/mozilla-lockbox/lockbox-android/blob/76145359c70508beef3fb0eb017011fbedbef00a/codecov.yml#L9-L10

sashei:
@mcomella apologies for the delay! I've modified the bitrise command for building jacoco; babysitting it now to get the branch merged :)

build.gradle
```diff
@@ -19,6 +19,9 @@ buildscript {
     repositories {
         google()
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
```
sashei: 👍

app/build.gradle
```diff
@@ -192,4 +193,42 @@ configurations.all { config ->
             }
         }
     }
-}
\ No newline at end of file
+}
+
+// Set up jacoco for code coverage.
+if (project.hasProperty("coverage")) {
```
sashei: ooo

sashei:
@mcomella woo, looks like I've got codecov reporting from your branch!! would you like to do anything else before I merge this?

thanks again 💯 🚀 🌻

mcomella:
@sashei I added the documentation commit. I hope that's a good place!

PR #496, opened 2019-03-11
Fixup no-username bugs (crash and string change)

### LINKED ISSUE #488: No username entry crashes with "null pointer item detail fragment"
Steps to reproduce: 

1. login
2. navigate to the item list screen.
3. Tap on a login with an empty username

Expected: 
An item detail screen including the username field as "(no username)". 

Observed: 
The app crashes with a `NullPointerException`.

jhugman:
Fixes #488.

This PR fixes a crash caused by `kotlinx.android` auto generated code being incorrectly generated and incorrectly used. This PR does not address the generated code, which allowed: 

```kotlin
inputUsername.btnUsernameCopy
```

to be used, which resolves to `null`.

It also addresses an issue in autofill which caused strings containing "null" in display strings for autofill.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/autofill/FillResponseBuilder.kt
+++ b/app/src/main/java/mozilla/lockbox/autofill/FillResponseBuilder.kt
@@ -97,17 +97,18 @@ open class FillResponseBuilder(
         val datasetBuilder = Dataset.Builder()
 
         val title = titleFromHostname(credential.hostname)
+        val username = credential.username ?: context.getString(R.string.no_username)
 
         parsedStructure.usernameId?.let {
             val presentation = RemoteViews(context.packageName, R.layout.autofill_item)
                 .apply {
-                    setTextViewText(R.id.autofillValue, credential.username)
+                    setTextViewText(R.id.autofillValue, username)
                     setTextViewText(R.id.hostname, title)
                 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/mozilla/lockbox/autofill/FillResponseBuilder.kt
```diff
@@ -97,25 +97,26 @@ open class FillResponseBuilder(
...
                 .apply {
-                    setTextViewText(R.id.autofillValue, credential.username)
+                    setTextViewText(R.id.autofillValue, username)
                     setTextViewText(R.id.hostname, title)
                 }
 
             datasetBuilder.setValue(
                 it,
-                AutofillValue.forText(credential.username),
+                AutofillValue.forText(credential.username ?: ""),
```
eliserichards: what happens when the `credential.username` is null?

jhugman: When the `credential.username` is `null`, the empty string is used instead.

This means that the input field would be filled with the empty string rather than "null". (i.e. `null.toString()`)

jhugman: @eliserichards I've added steps to reproduce in the original issue #488

PR #497, opened 2019-03-12
74: Adjust Integration

### LINKED ISSUE #74: Include Adjust integration
- [x] Integrate the Adjust SDK: should be in the APK module's build.gradle, see instructions [here](https://github.com/adjust/android_sdk#add-the-sdk-to-your-project)
- [x] Set up the Lockbox Android token: 383z4i46o48w
  - [x] Provide marketing with the app ID (mozilla.lockbox)

eliserichards:
Fixes #74 
Removed Google Play Services from `build.gradle` and Proguard settings. Filing issue #500 to confirm at a later time if we want to integrate the AdvertisingId for in-app-store marketing.

## Testing and Review Notes
Used the Google Play Referrer API instead of the Google Play Store intent as the intent is set to be deprecated.

[Logs showing](https://github.com/adjust/android_sdk#build-your-app) Adjust `SANDBOX` being used:
```
2019-03-12 12:45:44.446 22458-22458/mozilla.lockbox W/Adjust: SANDBOX: Adjust is running in Sandbox mode. Use this setting for testing. Don't forget to set the environment to `production` before publishing!
2019-03-12 12:45:45.337 22458-22500/mozilla.lockbox I/Adjust: Google Play Services Advertising ID read correctly at start time
```

Questions: 
* What is a good place to store the app token? I chose `Constant.App.`
* What level of logging would we like? [Default](https://github.com/adjust/android_sdk#adjust-logging) is `.INFO`. 

## To Do


app/build.gradle
```diff
@@ -115,6 +115,9 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-common-java8:2.1.0-alpha02'
     implementation "android.arch.navigation:navigation-fragment:$navigation_version"
     implementation "android.arch.navigation:navigation-ui-ktx:$navigation_version"
+    implementation 'com.adjust.sdk:adjust-android:4.17.0'
+    implementation 'com.android.installreferrer:installreferrer:1.0'
+    implementation 'com.google.android.gms:play-services-analytics:16.0.7'
```
sashei: 😞

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -54,6 +59,20 @@ open class LockboxApplication : Application() {
         injectContext()
         setupLifecycleListener()
         setupSentry()
+
+        // Adjust Integration
+        val appToken = Constant.App.appToken
+
+        // ONLY USE SANDBOX when we are developing locally or testing.
+        // ONLY USE PRODUCTION when it is a real play store build.
+        val environment = AdjustConfig.ENVIRONMENT_SANDBOX
```
sashei: Let's have this switch on the build variant rather than needing to do it manually here... I'm pretty sure this kind of thing is what the `isDebug()` helper method is for.

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
```diff
@@ -130,4 +149,24 @@ open class LockboxApplication : Application() {
...
         return device == "robolectric" && product == "robolectric"
     }
+
+    private class AdjustLifecycleCallbacks : ActivityLifecycleCallbacks {
+        override fun onActivityResumed(activity: Activity) {
+            Adjust.onResume()
+        }
+
+        override fun onActivityPaused(activity: Activity) {
+            Adjust.onPause()
```
sashei: Let's check with @sandysage about what's useful for marketing (do we need `onPause`?). If we're only tracking for the install / app open, we can just use `onActivityCreated`. 

Also, I'd like to see the `AdjustLifecycleCallbacks` class in its own file; the `LockboxApplication` can get bloated quickly!

app/build.gradle
eliserichards: why the sad face?? 🤔

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
eliserichards: `onResume()` and `onPause()` were part of the initial instructions to integrate this. I'm not sure if we _need_ to implement these or if they can be skipped.

app/build.gradle
sashei: google play services! :p

eliserichards:
@sashei The Adjust integration instructions (in Java) say that the class `AdjustLifecycleCallbacks : ActivityLifecycleCallbacks` should be `private static final`. In order to pull it out of the LockboxApplication, it won't be private anymore. Do you think this is ok?

app/src/main/java/mozilla/lockbox/LockboxApplication.kt
sashei: Yeah, I checked out those instructions.... I don't think we need to cargo-cult those pieces from the Adjust folks. @sandysage do you have an idea about what metrics we're trying to gather from the integration? Just installs? That will help guide which of these methods is going to be useful.

app/build.gradle
eliserichards: Should I take this out?

sashei: no, it's required for adjust. with it, however, we can't release into app stores other than the google play store (e.g. F-droid).

sashei:
@eliserichards let's keep it `static` if possible, but I think it's ok to drop the `private`!

devinreams:
Merging even though failing, due to flakey 502 on signing step. 😔

PR #499, opened 2019-03-13
Navigation/Rx sanity.

### LINKED ISSUE #415: List is empty even though logged in
I'm logged into lockbox with my account, but I'm seeing the "No Entries Found" screen (sounds similar to #360, but in fact nothing ever shows up).

### LINKED ISSUE #448: Back button from Welcome Screen takes you to "No Entries Found"
When launching Lockbox fresh, tapping the Back button takes the user to the "No Entires Found" screen instead of leaving the app.

## Rerequisites
* Lockbox is reset or first-time run

## Steps
1. Launch Lockbox
   - see Welcome screen
2. tap Back button

## Expected Results
The user exits Lockbox

## Actual Results
The user is taken to "No Entries Found"

### LINKED ISSUE #452: Implement an explicit default state in nav controller (show a blank or simple layout  vs "no entries list")
Currently the splash screen that displays Lokie is a variant of the entry list and not an explicit view. We're finding a few potentially related issues around "no entries found" displaying at seemingly odd times:

* #448
* #447
* #440 (maybe)
* #415 (maybe)

jhugman:
As navigation and rx are very well designed, they encapsulate their inner state very well. Unfortunately, this causes problems for our app: 

 * the app takes sometime to start up, and it's clear what the first thing the user should see in the meantime.
 * modelling the navigation back stack simplistically complicates this because we don't know how we should start.

Fixes #452
Fixes #448
Fixes #415

This PR does several things: 

 * adds a blank fragment whose only job is to be blank while a splash screen is displayed via a theme. This should be dismissed pretty quickly, but avoid a flash of an item list screen.
 * introduces a stack replay subject, with tests
 * uses the new subject to model the back stack of the application in a way not possible before. 
 * double checks that the `graph_main.xml` has all the correct XML files.
 * fixes `IllegalStateException`s being through by navigation controller.

This is not ready to land yet because: 

 * opening an item detail view and then opening the host in a browser, then clicking back twice should take you to the item list, instead, the second back is ignored. This is caused by:
 * as part of PR #485, we moved to subscribe and unsubscribe from the `RouteStore` eagerly, in `onPause` and `onResume`. This, I think, was a mistake. Instead, we should either make the autofill extension use its own RouteStore, use separate Autofil versions of FingerprintDialogAction, ensure only one subscriber is allowed at a time for the route store; or something else.
 * the `StackReplaySubject` is largely a cut and paste and change version of the `PublishSubject`. This could do with a tidy.

Asking for early feedback for this PR.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
@@ -29,10 +29,8 @@ open class RouteAction(
     data class OpenWebsite(val url: String) :
         RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.open_in_browser)
 
-    // This route shouldn't be used on its own. It is a terminator route to ensure that actions
-    // that leave the app do not get replayed once the app becomes foregrounded.
-    // It never flows through the dispatcher, so does not get wrongly recorded by telemetry.
-    object NoFollowOnReturn : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
+    // This should _only_ be triggered by pressing the back button.
+    object InternalBack : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
 
     data class SystemSetting(val setting: SettingIntent) :
         RouteAction(TelemetryEventMethod.show, TelemetryEventObject.settings_system)
--- a/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/RouteAction.kt
@@ -12,6 +12,7 @@ import androidx.annotation.RequiresApi
 import androidx.annotation.StringRes
 import mozilla.lockbox.R
 import mozilla.lockbox.flux.Action
+import mozilla.lockbox.support.Constant
 
 open class RouteAction(
     override val eventMethod: TelemetryEventMethod,
@@ -26,14 +27,10 @@ open class RouteAction(
     object LockScreen : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.lock_screen)
     object Filter : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.filter)
     data class ItemDetail(val id: String) : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.entry_detail)
-    data class OpenWebsite(val url: String) :
-        RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.open_in_browser)
+
 
     // This should _only_ be triggered by pressing the back button.
     object InternalBack : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
-
-    data class SystemSetting(val setting: SettingIntent) :
-        RouteAction(TelemetryEventMethod.show, TelemetryEventObject.settings_system)
 
     sealed class DialogFragment(
         @StringRes val dialogTitle: Int,
@@ -45,6 +42,12 @@ open class RouteAction(
         object AutofillSearchDialog : DialogFragment(R.string.autofill)
     }
 
+    open class SystemIntent(
+        val requestCode: Int = Constant.RequestCode.noResult,
+        eventMethod: TelemetryEventMethod,
+        eventObject: TelemetryEventObject
+    ) : RouteAction(eventMethod, eventObject)
+
     sealed class Onboarding(
         eventObject: TelemetryEventObject
     ) : RouteAction(TelemetryEventMethod.show, eventObject) {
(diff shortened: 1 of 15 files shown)


app/src/main/java/mozilla/lockbox/presenter/BackablePresenter.kt
```diff
@@ -1,15 +1,24 @@
...
-    // No methods yet.
+    val backButtonClicks: Observable<Unit>
 }
 
 class BackablePresenter(
     val view: BackableView,
     val dispatcher: Dispatcher = Dispatcher.shared
 ) : Presenter() {
-    // No methods yet.
+    override fun onViewReady() {
```
sashei: will you enforce that views inheriting from the `BackablePresenter` should call `super.onViewReady()`?

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -174,11 +180,19 @@ class RoutePresenter(
                     "This is a developer bug, fixable by adding an action to graph_main.xml"
             )
         } else {
-            val clearBackStack = src.getAction(transition)?.navOptions?.shouldLaunchSingleTop() ?: false
+
+            val action = src.getAction(transition)
```
sashei: get the action _before_ clearing the backstack

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -174,11 +180,19 @@ class RoutePresenter(
...
+            val clearBackStack = action?.navOptions?.shouldLaunchSingleTop() ?: false
             if (clearBackStack) {
                 while (navController.popBackStack()) {
                     // NOP
                 }
+                routeStore.clearBackStack()
+
+                action?.let {
```
sashei: why is this inside of the `if (clearBackStack)` block?

app/src/main/java/mozilla/lockbox/action/RouteAction.kt
```diff
@@ -29,10 +29,8 @@ open class RouteAction(
     data class OpenWebsite(val url: String) :
         RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.open_in_browser)
 
-    // This route shouldn't be used on its own. It is a terminator route to ensure that actions
-    // that leave the app do not get replayed once the app becomes foregrounded.
-    // It never flows through the dispatcher, so does not get wrongly recorded by telemetry.
-    object NoFollowOnReturn : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
+    // This should _only_ be triggered by pressing the back button.
+    object InternalBack : RouteAction(TelemetryEventMethod.tap, TelemetryEventObject.back)
```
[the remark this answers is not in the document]
jhugman: That sounds like a good idea.

After some exploration (it's not lots of change), it wasn't a flawless change, so suggest iterating this in another PR.

app/src/main/java/mozilla/lockbox/flux/StackReplaySubject.kt
```diff
@@ -0,0 +1,284 @@
...
+    override fun subscribeActual(t: Observer<in T>) {
+        val ps = StackDisposable(t, this)
+        t.onSubscribe(ps)
+        if (add(ps)) {
+            // if cancellation happened while a successful add, the remove() didn't work
+            // so we need to do it again
+            if (ps.isDisposed) {
+                remove(ps)
+            } else {
+                replay(ps)
```
jhugman: `replay` the latest top most on the stack when subscribing.

app/src/main/java/mozilla/lockbox/presenter/BackablePresenter.kt
jhugman: :+1:

jhugman: Good catch. 🛁 

Right now, `BackablePresenter` is _only_ used by `BackableFragment`s — i.e. the fragments have two presenters each.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: Yerk; you're right this could be cleaner.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jhugman: 👍

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -113,7 +120,7 @@ class RoutePresenter(
                     }
                 }
             }
-            .flatMapIterable { it }
+            .flatMapIterable { listOf(RouteAction.InternalBack) + it }
```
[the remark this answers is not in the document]
jhugman: Yeah, I think I'd prefer the dialogs and dialog fragments to dispatch their own `InternalBack` route actions.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -42,27 +43,33 @@ class RoutePresenter(
     private val settingStore: SettingStore = SettingStore.shared
 ) : Presenter() {
     private lateinit var navController: NavController
+    private lateinit var backListener: OnBackPressedCallback
```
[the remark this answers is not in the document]
jhugman: Moved to declaration.

sashei:
I'm still seeing this onboarding screen borked: 
image

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -133,27 +156,32 @@ class RoutePresenter(
...
+        try {
+            dialogFragment.setTargetFragment(currentFragment, 0)
+            dialogFragment.show(navHostFragmentManager, dialogFragment.javaClass.name)
+            dialogFragment.setupDialog(destination.dialogTitle, destination.dialogSubtitle)
+        } catch (e: IllegalStateException) {
+            log.error("Could not show dialog", e)
         }
     }
+
+    private fun showUnlockFallback(action: RouteAction.UnlockFallbackDialog) {
```
sashei: oh _awesome_

jhugman:
@sashei can't reproduce the onboarding borkage, fixed in 7021fdb73d58d561326576a4b05528dcbff30814. Have you any STR?

PR #502, opened 2019-03-14
appropriate behavior when backgrounding autofill + locking

### LINKED ISSUE #489: Backgrounding Autofill for some period > autolock time will launch the fingerprint dialog again
Steps to reproduce:
- Launch into the Search view in autofill (either unlocking or from Search Your Entries row)
- Use the home button or simply let the device go to sleep for > 5minutes (or whatever the autolock time is)

**expected:** the search dialog goes away, only the fingerprint dialog remains
**actual:** a second fingerprint dialog is launched over the search dialog. unlocking brings you to a second search dialog.

sashei:
Fixes #489

## Testing and Review Notes

The best way that I found to test this was to try to autofill a site for which we have no credentials. I would subsequently enter something into the search box so that it's expanded with some results, and then put the device to sleep for the autolock time. Ideally, upon unlocking the device, the search dialog goes away and the fingerprint or PIN screen is shown.

## To Do

- [x] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/AutofillFilterBackdropPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/AutofillFilterBackdropPresenter.kt
@@ -1,0 +1,28 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.presenter
+
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.addTo
+import mozilla.lockbox.action.RouteAction
+import mozilla.lockbox.flux.Dispatcher
+import mozilla.lockbox.flux.Presenter
+import java.util.concurrent.TimeUnit
+
+interface AutofillFilterBackdropView
+
+class AutofillFilterBackdropPresenter(
+    val view: AutofillFilterBackdropView,
+    private val dispatcher: Dispatcher = Dispatcher.shared
+) : Presenter() {
+    override fun onViewReady() {
+        Observable.just(RouteAction.DialogFragment.AutofillSearchDialog)
+            .delay(1, TimeUnit.SECONDS)
+            .subscribe(dispatcher::dispatch)
+            .addTo(compositeDisposable)
+    }
+}
--- a/app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
@@ -9,6 +9,8 @@ import android.view.autofill.AutofillManager
 import androidx.annotation.IdRes
 import androidx.annotation.RequiresApi
 import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
 import androidx.navigation.NavController
 import androidx.navigation.Navigation
 import io.reactivex.Observable
@@ -44,6 +46,18 @@ class AutofillRoutePresenter(
     private val pslSupport: PublicSuffixSupport = PublicSuffixSupport.shared
 ) : Presenter() {
     private lateinit var navController: NavController
+
+    private val navHostFragmentManager: FragmentManager
+        get() {
+            val fragmentManager = activity.supportFragmentManager
+            val navHost = fragmentManager.fragments.last()
+            return navHost.childFragmentManager
+        }
+
+    private val currentFragment: Fragment
+        get() {
+            return navHostFragmentManager.fragments.last()
+        }
 
     override fun onViewReady() {
         navController = Navigation.findNavController(activity, R.id.autofill_fragment_nav_host)
@@ -82,10 +96,16 @@ class AutofillRoutePresenter(
 
     private fun route(action: RouteAction) {
         when (action) {
-            is RouteAction.LockScreen -> navigateToFragment(R.id.fragment_locked)
-            is RouteAction.ItemList -> showDialogFragment(AutofillFilterFragment(),
-                RouteAction.DialogFragment.AutofillSearchDialog
-            )
+            is RouteAction.LockScreen -> {
+                dismissDialogIfPresent()
+                navigateToFragment(R.id.fragment_locked)
+            }
+            is RouteAction.ItemList -> {
+                dismissDialogIfPresent()
+                navigateToFragment(R.id.fragment_filter_backdrop)
+            }
+            is RouteAction.DialogFragment.AutofillSearchDialog ->
+                showDialogFragment(AutofillFilterFragment(), action)
             is RouteAction.DialogFragment.FingerprintDialog ->
                 showDialogFragment(FingerprintAuthDialogFragment(), action)
         }
@@ -101,26 +121,33 @@ class AutofillRoutePresenter(
 
         val transition = findTransitionId(srcId, destinationId) ?: destinationId
 
-        if (transition == destinationId) {
+        val navOptions = if (transition == destinationId) {
             // Without being able to detect if we're in developer mode,
             // it is too dangerous to RuntimeException.
             val from = activity.resources.getResourceName(srcId)
             val to = activity.resources.getResourceName(destinationId)
+            val graphName = activity.resources.getResourceName(navController.graph.id)
             log.error(
                 "Cannot route from $from to $to. " +
-                    "This is a developer bug, fixable by adding an action to graph_autofill.xml"
+                    "This is a developer bug, fixable by adding an action to $graphName.xml and/or ${javaClass.simpleName}"
             )
+            null
         } else {
-            val clearBackStack = src.getAction(transition)?.navOptions?.shouldLaunchSingleTop() ?: false
-            if (clearBackStack) {
-                while (navController.popBackStack()) {
-                    // NOP
+            // Get the transition action out of the graph, before we manually clear the back
+            // stack, because it causes IllegalArgumentExceptions.
+            src.getAction(transition)?.navOptions?.let { navOptions ->
+                if (navOptions.shouldLaunchSingleTop()) {
+                    while (navController.popBackStack()) {
+                        // NOP
+                    }
+                    routeStore.clearBackStack()
                 }
+                navOptions
             }
         }
 
         try {
-            navController.navigate(transition, args)
+            navController.navigate(destinationId, args, navOptions)
         } catch (e: IllegalArgumentException) {
             log.error("This appears to be a bug in navController", e)
             navController.navigate(destinationId, args)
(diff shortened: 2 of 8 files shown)


app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -146,6 +156,14 @@ class AutofillRoutePresenter(
         }
     }
 
+    private fun <T : DialogFragment> dismissDialogIfPresent(clazz: Class<T>) {
+        val fragmentManager = activity.supportFragmentManager
+        val presentedDialog = fragmentManager.findFragmentByTag(clazz.name)
+        log.info("presented dialog: $presentedDialog")
```
jhugman: In general, I don't like that the route needs to have fore-knowledge of the dialog that we're moving from.

Additionally, I'm not entirely sure if this is only working accidentally. I think I'd prefer to see a 

```kotlin
<T : Fragment> fun FragmentManager.findChildFragmentByClass(clazz: Class<T>)
```

which looks at `fragment.childFragmentManager` of each fragment (currently, `showDialogFragment` uses the `NavHostFragment` as the `currentFragment`).

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -111,12 +115,18 @@ class AutofillRoutePresenter(
...
             if (clearBackStack) {
                 while (navController.popBackStack()) {
                     // NOP
                 }
             }
+
+            action?.let {
+                navController.navigate(it.destinationId, args, it.navOptions)
+                return
+            }
```
jhugman: The sooner we rationalize `AutofillRoutePresenter` and `RoutePresenter` the better.

sashei: https://github.com/mozilla-lockbox/lockbox-android/issues/508

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -128,10 +138,10 @@ class AutofillRoutePresenter(
     }
 
     private fun findTransitionId(@IdRes from: Int, @IdRes to: Int): Int? {
-        return when (Pair(from, to)) {
-            Pair(R.id.fragment_locked, R.id.fragment_filter) -> R.id.action_locked_to_filter
-            Pair(R.id.fragment_null, R.id.fragment_filter) -> R.id.action_to_filter
-            Pair(R.id.fragment_null, R.id.fragment_locked) -> R.id.action_to_locked
+        return when (from to to) {
```
eliserichards: I would rename `to`. Suggestions: `start to end`, `begin to end`, `from to dest`/`destination`...

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
sashei: I don't know if I'd call this "accidental" but I would call it "hacky" ;)

it relies on us setting the tag for the dialog fragments to their respective class names as obtained by `clazz.name`.... I'm open to other solutions tho.

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -101,51 +121,61 @@ class AutofillRoutePresenter(
...
+            dialogFragment.show(navHostFragmentManager, dialogFragment.javaClass.name)
             dialogFragment.setupDialog(destination.dialogTitle, destination.dialogSubtitle)
         } catch (e: IllegalStateException) {
             log.error("Could not show dialog", e)
         }
     }
 
+    private fun dismissDialogIfPresent() {
+        (currentFragment as? DialogFragment)?.dismiss()
+    }
```
jhugman: ✅ ✅ ✅

app/src/test/java/mozilla/lockbox/presenter/AutofillRoutePresenterTest.kt
```diff
@@ -86,6 +88,15 @@ class AutofillRoutePresenterTest {
...
     val fragmentManager: FragmentManager = mock(FragmentManager::class.java)
 
+    @Mock
+    val childFragmentManager: FragmentManager = mock(FragmentManager::class.java)
+
+    @Mock
+    val currentFragment: DialogFragment = mock(DialogFragment::class.java)
+
+    @Mock
+    val navHost: Fragment = mock(Fragment::class.java)
```
jhugman: Ah great, this is good to see tested.

app/src/main/res/navigation/graph_main.xml
```diff
@@ -9,9 +9,9 @@
             xmlns:app="http://schemas.android.com/apk/res-auto"
             xmlns:tools="http://schemas.android.com/tools"
             android:id="@+id/graph_main"
-            app:startDestination="@id/fragment_null">
+            app:startDestination="@id/fragment_filter_backdrop">
     <fragment
-            android:id="@+id/fragment_null"
+            android:id="@+id/fragment_filter_backdrop"
```
jhugman: I think that this is a little over zealous refactoring by Android Studio.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -230,9 +230,9 @@ class RoutePresenter(
         // If a RouteAction is called from a place the graph doesn't know about then
         // the app will log.error.
         return when (from to to) {
-            R.id.fragment_null to R.id.fragment_item_list -> R.id.action_init_to_unlocked
-            R.id.fragment_null to R.id.fragment_locked -> R.id.action_init_to_locked
-            R.id.fragment_null to R.id.fragment_welcome -> R.id.action_init_to_unprepared
+            R.id.fragment_filter_backdrop to R.id.fragment_item_list -> R.id.action_init_to_unlocked
+            R.id.fragment_filter_backdrop to R.id.fragment_locked -> R.id.action_init_to_locked
+            R.id.fragment_filter_backdrop to R.id.fragment_welcome -> R.id.action_init_to_unprepared
```
jhugman: I think that this is a little over zealous refactoring by Android Studio.

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -101,51 +121,61 @@ class AutofillRoutePresenter(
...
 
-    private fun findTransitionId(@IdRes from: Int, @IdRes to: Int): Int? {
-        return when (Pair(from, to)) {
-            Pair(R.id.fragment_locked, R.id.fragment_filter) -> R.id.action_locked_to_filter
-            Pair(R.id.fragment_null, R.id.fragment_filter) -> R.id.action_to_filter
-            Pair(R.id.fragment_null, R.id.fragment_locked) -> R.id.action_to_locked
+    private fun findTransitionId(@IdRes from: Int, @IdRes destination: Int): Int? {
+        return when (from to destination) {
+            R.id.fragment_locked to R.id.fragment_filter_backdrop -> R.id.action_to_filter
+            R.id.fragment_filter_backdrop to R.id.fragment_locked -> R.id.action_to_locked
```
jhugman: Nit: use `src` and `dest` here. `from` and `destination` is mixing your metaphors, like comparing `chalk` and `oranges`.

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
```diff
@@ -82,10 +96,16 @@ class AutofillRoutePresenter(
...
-            is RouteAction.LockScreen -> navigateToFragment(R.id.fragment_locked)
-            is RouteAction.ItemList -> showDialogFragment(AutofillFilterFragment(),
-                RouteAction.DialogFragment.AutofillSearchDialog
-            )
+            is RouteAction.LockScreen -> {
+                dismissDialogIfPresent()
+                navigateToFragment(R.id.fragment_locked)
+            }
+            is RouteAction.ItemList -> {
+                dismissDialogIfPresent()
```
jhugman: I'm still not altogether happy with `dismissDialogIfPresent`. It feels like the dialog is not cleaning up after its own mess.

app/src/main/res/navigation/graph_autofill.xml
```diff
@@ -9,17 +9,12 @@
             xmlns:app="http://schemas.android.com/apk/res-auto"
             xmlns:tools="http://schemas.android.com/tools"
             android:id="@+id/graph_autofill"
-            app:startDestination="@id/fragment_null">
+            app:startDestination="@id/fragment_filter_backdrop">
```
jhugman: This looks like it should cause problems when the app is locked. 

I think I'd prefer to keep using blank and then move on to the `AutofillFilterBackdropFragment`. This makes it congruent with the `graph_main.xml`.

app/src/main/java/mozilla/lockbox/presenter/AutofillFilterBackdropPresenter.kt
```diff
@@ -0,0 +1,28 @@
...
+
+interface AutofillFilterBackdropView
+
+class AutofillFilterBackdropPresenter(
+    val view: AutofillFilterBackdropView,
+    private val dispatcher: Dispatcher = Dispatcher.shared
+) : Presenter() {
+    override fun onViewReady() {
+        Observable.just(RouteAction.DialogFragment.AutofillSearchDialog)
+            .delay(1, TimeUnit.SECONDS)
```
eliserichards: What is the purpose of this `delay`?

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
eliserichards: Can this be handled in the navigation instead of here? It doesn't seem to make sense.

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
eliserichards: Why isn't the `dialog` set up to be dismissed on its own? Should we be doing this same thing for the `RoutePresenter`'s `showDialogFragment`? https://github.com/mozilla-lockbox/lockbox-android/blob/ac8add36bd4c79ad0b6db11881b5da928a2bbaa2/app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt#L166

app/src/main/java/mozilla/lockbox/view/AutofillFilterFragment.kt
```diff
@@ -84,9 +85,12 @@ class AutofillFilterFragment : DialogFragment(), AutofillFilterView {
         adapter.displayNoEntries(enabled)
     }
 
-    override fun onDestroyView() {
+    override fun onCancel(dialog: DialogInterface) {
         (onDismiss as PublishSubject).onNext(Unit)
+        super.onCancel(dialog)
```
eliserichards: Would ``` (currentFragment as? DialogFragment)?.dismiss()``` relavent here?

app/src/main/java/mozilla/lockbox/presenter/AutofillFilterBackdropPresenter.kt
sashei: It ensures that the correct fragment is presented before the `DialogFragment` gets presented / attached

app/src/main/java/mozilla/lockbox/presenter/AutofillRoutePresenter.kt
sashei: We don't have the same situation in the normal app; we rarely have to dismiss a dialog _without_ anything being tapped on it.

PR #552, opened 2019-03-28
Show VersionCode (build number) on settings screen

beatbrot:
Fixes #551

## Testing and Review Notes

Just open the settings and see the versioncode

## Screenshots or Videos

_(Optional: to clearly demonstrate the feature or fix to help with testing and reviews)_

## To Do

- [x] add “WIP” to the PR title if pushing up but not complete nor ready for review
- [ ] add unit tests
  - optional: consider adding instrumentation (integration/UI) tests
- [x] double check the original issue to confirm it is fully satisfied
- [x] add testing notes and screenshots in PR description to help guide reviewers
- [ ] consider running this branch in a debug simulator and check for memory leak notifications or any warnings
- [ ] request the "UX" team perform a design review (if/when applicable)
- [x] make sure CI builds are passing (e.g.: fix lint and other errors)
- [x] check on the [accessibility](https://mozilla-lockbox.github.io/lockbox-android/accessibility/) of any added UI

PR #562, opened 2019-03-29
Bump version to 1.0.1

### LINKED ISSUE #550: Bump version to 1.0.1
And updated the release docs if not in our instructions

devinreams:
Fixes #550

See also: #552 (watch for final decisions, merge against, which adds the build number)

## Testing and Review Notes

(user facing)

- open the app
- open the settings screen
- observe the version

## Screenshots or Videos

## To Do

- [x] add unit tests (N/A)
- [x] request the "UX" team perform a design review (N/A)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,8 +22,8 @@ android {
         minSdkVersion 24
         //noinspection OldTargetApi
         targetSdkVersion 26
-        versionCode 1
-        versionName "1.0"
+        versionCode 1.0.1
+        versionName "1.0.1"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         // The following argument makes the Android Test Orchestrator run its
         // "pm clear" command after each test invocation. This command ensures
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -23,7 +23,7 @@ android {
         //noinspection OldTargetApi
         targetSdkVersion 26
         versionCode 1
```
sashei: plz update versionCode as well! this is what the system uses to decide whether we need an upgrade.

I am seeing that this does not accept values of the form `1.0.1` however; maybe worth revisiting our numbering scheme?? Or we can use increasing values not related to the `versionName` for this value.

devinreams: >I am seeing that this does not accept values of the form 1.0.1 however

Right, because this should be an integer as I understand it. Like 3171 was the versionCode for versionName 1.0.0 (but versionCode is set by bitrise).

Perhaps I take the expected bitrise build number that would run when the next commit happens?

sashei: Yeah that sounds reasonable!

PR #568, opened 2019-04-01
autolock on device restart

### LINKED ISSUE #510: Restarting device after Lockbox automatically locks will unlock the app

sashei:
Fixes #510

## Testing and Review Notes

bug description is sufficient to describe the problem; the expected behavior is that the app will _always_ be locked after restarting the device (unless the Autolock time is `Never`)

the boot notification gets sent ~1 minute after the device is started up, so there is a brief window in which the lock screen can still be bypassed. however, the notification is always sent after the user has authenticated at a device level (PIN only) for the first time after a restart, so it's unlikely that Lockbox security would prevent accessing passwords in this case anyway.

## To Do

- [x] add unit tests

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,7 +59,7 @@
                 <action android:name="com.android.vending.INSTALL_REFERRER" />
             </intent-filter>
         </receiver>
-        
+
         <receiver android:name=".BootReceiver">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED"/>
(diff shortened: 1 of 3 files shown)


sashei:
@linuxwolf what is your feeling about the boot notification delay? @eliserichards and I have been chatting and think that it's unlikely users will go _immediately_ to Lockbox (do I have to start using Lockwise yet??) after starting their devices and thus we don't have to be overly concerned with that case.

A possible workaround is to _additionally_ lock the app whenever the system clock reads <5min, but this seems brittle to extremely slow devices.

linuxwolf:
I think relying solely on `BOOT_COMPLETED` is safe enough.  I agree that it's unlikely someone will jump straight to Lock*, and it is extremely likely that person could get into Lock* anyway.

If it does become an issue, we enhance in the future with additional constraints.

PR #587, opened 2019-04-08
document autofill metrics

irrationalagent:
Hi Sasha, can you clarify any of my misconceptions about what the autofill events measure, or if there was a better way to describe them? I'm pretty sketchy on their meaning and it will help me narrow down my queries to only count those events that are involved with the user actually trying to use autofill. 

(anyone else can jump in here as well, but I think it was sasha who implemented the events)

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -168,13 +168,13 @@ https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/c
 	* `category`: action
 	* `method`:
 		* autofill_error (there was an error when attempting to fill a form),
-		* autofill_single (the autofill API returned one credential that fit a form),
-		* autofill_multiple (the autofill API returned multiple possible matches to a form),
+		* autofill_multiple (the autofill API returned a list possible credentials to fill a form),
+		* autofill_single (a credential was selected from the list of possible credentials to fill a form),
 		* autofill_filter (the user filtered a returned list of autofill credentials),
 		* autofill_locked (the datastore was locked when a form triggered the autofill API),
 		* autofill_cancel (the user cancelled form autofill),
-		* background (always when the app is backgrounded?),
-		* foreground (always when the app is foregrounded?)
+		* background (when the autofill service is backgrounded, doesn't necessarily involve user-facing UX),
+		* foreground (always when the autofill service is foregrounded, doesn't necessarily involve user-facing UX)
 	* `object`: autofill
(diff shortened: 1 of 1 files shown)


irrationalagent:
oops, didn't realize that sasha was out. can wait until she gets back, or if someone else wants to take it.

docs/metrics.md
```diff
@@ -164,6 +164,21 @@ https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/c
...
+13. When a user interacts with the autofill functionality
+	* `category`: action
+	* `method`:
+		* autofill_error (there was an error when attempting to fill a form),
+		* autofill_single (the autofill API returned one credential that fit a form),
+		* autofill_multiple (the autofill API returned multiple possible matches to a form),
+		* autofill_filter (the user filtered a returned list of autofill credentials),
+		* autofill_locked (the datastore was locked when a form triggered the autofill API),
+		* autofill_cancel (the user cancelled form autofill),
+		* background (always when the app is backgrounded?),
```
linuxwolf: technically, whenever the autofill service is backgrounded/foregrounded ... which can be often.

eliserichards:
**Foreground/Background:** Whenever there is an autofill request, the service is connected and `TelemetryEventMethod.foreground` is called. Once the request is processed, we disconnect and call `TelemetryEventMethod.background`. This is kind of like a "session", if that makes sense. 

**Complete/CompleteMultiple:** `AutofillAction.Complete` (aka `TelemetryEventMethod.autofill_single`) is also used when we're filtering and an item is selected. `AutofillAction.CompleteMultiple` (aka `TelemetryEventMethod.autofill_multiple`) happens in the default autofill workflows (so this can be called even if there is only one credential).

docs/metrics.md
```diff
@@ -164,6 +164,21 @@ https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/c
...
+		* foreground (always when the app is foregrounded?)
```
sashei: the foreground and background events will be triggered even in cases where the autofill service does not find any fields to fill, and therefore does not present any lockbox UI to the user. there are probably lots of forms (eg, credit cards) that the system "requests" lockbox on, that we don't subsequently fill. the numbers of these events might be interesting in that context but don't reflect overall autofill engagement in the same way as `autofill-` `_single`, `_multiple`, `_filter`, etc.

PR #590, opened 2019-04-11
589-UItests-update

### LINKED ISSUE #589: UITests - some UI tests are failing
The tests are great and there are already a lot! which is really nice :) 
I have been running them to get familiar with the environment and in order to know the status before enabling them (if applies) in CI. I'm getting some failures:
- testOnboardingConfirmation
- testItemList
- fingerprintSuccessNavigatesToItemList
- fingerprintSkipButtonNavigatesToItemList
- class LockboxAutofillServiceTest

You may be already aware of them but I am opening the issue to track their status and fix in which I could start working on if you want.
The first two are easy to fix by modifying the test and navigator, same for the next two but need to check if there is an issue there, in order to have them both green, they have to follow the same exact steps in the test...
For the last one, there is not any test in the test suite itself, it fails with an initialization error. Not sure how to fix it yet but will investigate.

isabelrios:
Fixes #589 
This PRs tries to update some of the tests to have them running regularly. There are some tests disabled atm due to more work/investigation needed to have them passing under all scenarios.

In summary:
- AppRoutePresenterTest: three tests were failing. Fixed by updating the route and since tests run on simulator, adding an extra step atm to tap on Skip when the Secure your device alert is shown
- AutoLockTests: disabled. Running ok on device but not on simulator. This will require some more work.
- OnboardingTests: disabled four tests that were failing on simulator. 
- LockboxAutofillServiceTest: disabled due to Inizialization error. Not sure about this Test suite since it does not actually contain any test but it fails in both device and sim.
- ItemsDetailTest: showToastWhenUsernameCopied test was failing because the selecting item did not have username. That's why the navigator has been changed to take the second entry in the list.

### DIFF
--- a/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
@@ -91,7 +91,9 @@ class Navigator {
 
     fun gotoFxALogin() {
         welcome { tapGetStarted() }
-        welcome { tapSkipSecureYourDevice() }
+        if (!FingerprintStore.shared.isDeviceSecure) {
+            welcome { tapSkipSecureYourDevice() }
+        }
         checkAtFxALogin()
     }
 
@@ -112,8 +114,10 @@ class Navigator {
     fun gotoAutofillOnboarding() {
         gotoFxALogin()
         fxaLogin { tapPlaceholderLogin() }
-        checkAtFingerprintOnboarding()
-        fingerprintOnboardingScreen { tapSkip() }
+        if (FingerprintStore.shared.isDeviceSecure) {
+            checkAtFingerprintOnboarding()
+            fingerprintOnboardingScreen { tapSkip() }
+        }
         checkAtAutofillOnboarding()
     }
 
--- a/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/Navigator.kt
@@ -91,6 +91,7 @@ class Navigator {
 
     fun gotoFxALogin() {
         welcome { tapGetStarted() }
+        welcome { tapSkipSecureYourDevice() }
         checkAtFxALogin()
     }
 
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/mozilla/lockbox/Navigator.kt
```diff
@@ -91,6 +91,7 @@ class Navigator {
 
     fun gotoFxALogin() {
         welcome { tapGetStarted() }
+        welcome { tapSkipSecureYourDevice() }
```
sashei: I wonder if we can add some logic to make this only tap `skip` when the device doesn't have any security... I think the `FingerprintStore.shared.isDeviceSecure` is available from the UI tests

isabelrios: Oh I did not know that...I just tried and yup, looks like that works! I will update this PR with that change. That allows to enable two more existing tests :)

isabelrios:
> This is looking really nice, thanks for spending some time getting these tests cleaned up @isabelrios !!

Oh it was easy! you did a great job with the tests! they are very clear to read and understand!

app/src/androidTest/java/mozilla/lockbox/Navigator.kt
```diff
@@ -91,6 +91,9 @@ class Navigator {
 
     fun gotoFxALogin() {
         welcome { tapGetStarted() }
+        if (!FingerprintStore.shared.isDeviceSecure) {
```
isabelrios: @sashei is it being used correctly here? At least all test are passing..but just to be sure...

app/src/androidTest/java/mozilla/lockbox/Navigator.kt
```diff
@@ -111,8 +114,10 @@ class Navigator {
     fun gotoAutofillOnboarding() {
         gotoFxALogin()
         fxaLogin { tapPlaceholderLogin() }
-        checkAtFingerprintOnboarding()
-        fingerprintOnboardingScreen { tapSkip() }
+        if (FingerprintStore.shared.isDeviceSecure) {
```
isabelrios: @sashei and same here...

sashei:
looks great @isabelrios !!

PR #600, opened 2019-04-18
Update app services and android components, and initialize megazord with networking

thomcc:
Doing this to help y'all along for integrating networking. Can't land until a-c does a non-snapshot release of 0.51.0.

Someone else should possibly take over this though, since unless it can land more or less as-is (with the removal of the -SNAPSHOT and the maven repo) I might not have time.

## To Do

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@
 buildscript {
     ext.android_support_version = '28.0.0'
     ext.kotlin_version = '1.3.0'
-    ext.android_components_version = '0.40.0'
+    ext.android_components_version = '0.51.0-SNAPSHOT'
     ext.lifecycle_version = '1.1.1'
     ext.navigation_version = '1.0.0'
     ext.rxbinding_version = '2.2.0'
@@ -23,6 +23,7 @@ buildscript {
     repositories {
         google()
         jcenter()
+
         maven {
             url "https://plugins.gradle.org/m2/"
         }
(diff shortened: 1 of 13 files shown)


build.gradle
```diff
@@ -9,7 +9,7 @@
 buildscript {
     ext.android_support_version = '28.0.0'
     ext.kotlin_version = '1.3.0'
-    ext.android_components_version = '0.40.0'
+    ext.android_components_version = '0.51.0-SNAPSHOT'
```
devinreams: >may have something to say about merge dates, since this PR upgrades @mozilla-mobile/android-components to a snapshot version.

Thanks @jhugman. Yes, good point. We quickly discussed waiting and I believe 0.51.0 should go out as a final [release](https://github.com/mozilla-mobile/android-components/releases) today? So we can wait for that, bump this, then re-test and merge 🤞

thomcc: Unfortunately, I found out yesterday that, IIUC, no a-c release is going out this week, https://mozilla.slack.com/archives/CA4JKKLSG/p1555962016115900

You can probably take it up with that team, and they might change their mind, I didn't press the issue too much when they told me.

devinreams: >Unfortunately, I found out yesterday that, IIUC, no a-c release is going out this week

AHA! Thanks for the context @thomcc.

@eliserichards @sashei is this OK by you to keep un-merged until next week? Or would you like to merge now and build/test against it for a few days and then file a follow-up to jump off the "SNAPSHOT" for later next week?

sashei: I'm fine keeping it un-merged until the next a-c release.

PR #605, opened 2019-04-19
Bump to version 1.0.2

devinreams:
Will wait to confirm the bitrise build number is set and appears as expected (it's "0", also as expected, when testing locally).

image


devinreams:
Confirmed with build 3426

PR #620, opened 2019-04-24
619-ui-tests-bitrise

isabelrios:
This PR is to try to have the UI tests working on Bitrise. I had to create a branch directly here instead of doing that from my fork so that it can run the Bitrise testing workflow where tests are enabled.

Since the tests start running but fail after the second one and before launching the third one let's comment those to see if there is a problem with them on Bitrise. Locally both, and all tests work.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,7 +88,7 @@ dependencies {
     implementation "androidx.recyclerview:recyclerview:$recyclerview_version"
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'androidx.exifinterface:exifinterface:1.0.0'
-    implementation 'androidx.constraintlayout:constraintlayout:2.0.0-alpha4'
+    implementation "androidx.constraintlayout:constraintlayout:2.0.0-alpha5"
     implementation "com.jakewharton.rxbinding2:rxbinding:${rxbinding_version}"
     implementation "com.jakewharton.rxbinding2:rxbinding-design-kotlin:${rxbinding_version}"
     implementation "com.jakewharton.rxbinding2:rxbinding-appcompat-v7-kotlin:${rxbinding_version}"
--- a/app/src/androidTest/java/mozilla/lockbox/AppRoutePresenterTest.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/AppRoutePresenterTest.kt
@@ -115,12 +115,11 @@ open class AppRoutePresenterTest {
         navigator.checkAtItemList()
     }
 
-    @Ignore
     @Test
     fun testNoSecurityDialog() {
-//        navigator.gotoNoSecurityDialog()
-//        navigator.back(true)
-//        navigator.checkAtItemList()
+        navigator.gotoNoSecurityDialog()
+        navigator.back(true)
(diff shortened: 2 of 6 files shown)


isabelrios:
Hey @sashei! After some work here I got 17 (out of 30) [tests working](https://app.bitrise.io/build/1a73105292113d56) on Bitrise. I would like to ask you for review so that I get some feedback on the fixes. Some of them are not very nice, so please let me know if you think there is a better way to enable them.
Thanks!

app/src/androidTest/java/mozilla/lockbox/AppRoutePresenterTest.kt
```diff
@@ -111,11 +117,12 @@ open class AppRoutePresenterTest {
 
     @Test
     fun testNoSecurityDialog() {
-//        navigator.gotoNoSecurityDialog()
-//        navigator.back(true)
-//        navigator.checkAtItemList()
+        navigator.gotoNoSecurityDialog()
```
sashei: 🙌

app/src/androidTest/java/mozilla/lockbox/OnboardingTest.kt
```diff
@@ -49,8 +56,11 @@ open class OnboardingTest {
         navigator.gotoAutofillOnboarding()
         autofillOnboardingScreen { tapSkip() }
         navigator.checkAtOnboardingConfirmation()
+        onboardingConfirmationScreen { clickFinish() }
+        navigator.disconnectAccount()
```
sashei: do you need to have this both in the test and in the `BeforeTest`?

app/src/androidTest/java/mozilla/lockbox/AppRoutePresenterTest.kt
```diff
@@ -93,8 +96,11 @@ open class AppRoutePresenterTest {
         navigator.gotoDisconnectDisclaimer()
         navigator.back()
         navigator.checkAtAccountSetting()
+        navigator.back()
+        navigator.back()
```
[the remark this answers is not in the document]
isabelrios: It is like for the simulator on Firebase we need to go back from the Disconnect message and from Account settings so that the next tests work...strange :(

app/src/androidTest/java/mozilla/lockbox/OnboardingTest.kt
isabelrios: you are right, we don't need this here :)

app/build.gradle
```diff
@@ -88,7 +88,7 @@ dependencies {
     implementation "androidx.recyclerview:recyclerview:$recyclerview_version"
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'androidx.exifinterface:exifinterface:1.0.0'
-    implementation 'androidx.constraintlayout:constraintlayout:2.0.0-alpha4'
+    implementation "androidx.constraintlayout:constraintlayout:2.0.0-alpha5"
```
isabelrios: I have to add this so that there is not any lint error, @sashei is that fine? would not like to merge the PR without approval on this change (after the review...) Thanks!

PR #623, opened 2019-04-24
597: DumpNode crashes on Android Q

### LINKED ISSUE #597: DumpNode crashes on Android Q
When running debug builds on Android Q, and a fillable field cannot be found, the `ViewNode.dump()` extension method throws an `IllegalStateException`.  This then prevents the autofill service from invoking the callback and therefore times out.

From Sentry:

```
IllegalStateException: value must be a text value, not type=2
    at mozilla.lockbox.extensions.ViewNode_Kt.dumpNode(ViewNode+.kt:23)
    at mozilla.lockbox.extensions.ViewNode_Kt.dumpChildren(ViewNode+.kt:58)
    at mozilla.lockbox.extensions.ViewNode_Kt.dumpNode(ViewNode+.kt:42)
    at mozilla.lockbox.extensions.ViewNode_Kt.dumpChildren(ViewNode+.kt:58)
    at mozilla.lockbox.extensions.ViewNode_Kt.dumpNode(ViewNode+.kt:42)
...
(21 additional frame(s) were not displayed)
```

https://sentry.prod.mozaws.net/operations/lockbox-android/issues/5550582/?query=is:unresolved

eliserichards:
Fixes #597

## To Do

- [ ] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
+++ b/app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
@@ -4,8 +4,6 @@ import android.annotation.TargetApi
 import android.app.assist.AssistStructure
 import android.os.Build
 
-private const val emptyString = ""
-
 @TargetApi(Build.VERSION_CODES.O)
 fun AssistStructure.ViewNode.dump(): String {
     val sb = StringBuilder("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
@@ -14,39 +12,23 @@ fun AssistStructure.ViewNode.dump(): String {
 
 @TargetApi(Build.VERSION_CODES.O)
 private fun AssistStructure.ViewNode.dumpNode(sb: StringBuilder = StringBuilder()): StringBuilder {
+
     val name = htmlInfo?.tag ?: className?.split('.')?.last() ?: "unknown"
 
-    var attrs: List<Pair<String, String>> = listOf(
+    val viewAttrs = listOf(
         "idEntry" to idEntry,
         "idPackage" to idPackage,
         "idType" to idType,
         "webDomain" to webDomain,
         "hint" to hint,
-        "autofillValue" to
-            if (autofillValue != null && autofillValue?.isText!!) {
-                autofillValue!!.textValue.toString()
-            } else {
-                emptyString
-            },
-        "autofillHints" to
-            if (autofillHints.isNullOrEmpty()) {
-                emptyString
-            } else {
-                autofillHints!!.joinToString(", ")
-            },
-        "autofillOptions" to
-            if (autofillOptions.isNullOrEmpty()) {
-                emptyString
-            } else {
-                autofillOptions!!.joinToString(", ")
-            }
+        "autofillValue" to autofillValue?.let { if (it.isText) it.textValue.toString() else null },
+        "autofillHints" to autofillHints?.joinToString(", "),
+        "autofillOptions" to autofillOptions?.joinToString(", ")
     )
 
-    htmlInfo?.attributes?.let { attributes ->
-        attrs += attributes.map { it.first to it.second }
-    }
+    val htmlAttrs = htmlInfo?.attributes?.map { it.first to it.second } ?: emptyList()
 
-    attrs = attrs.filter { it.first != emptyString && it.second != "null" }
+    val attrs = (viewAttrs + htmlAttrs).filter { it.second != null }
 
     sb.append("<$name")
     if (!attrs.isEmpty()) {
--- a/app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
+++ b/app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
@@ -16,22 +16,37 @@ fun AssistStructure.ViewNode.dump(): String {
 private fun AssistStructure.ViewNode.dumpNode(sb: StringBuilder = StringBuilder()): StringBuilder {
     val name = htmlInfo?.tag ?: className?.split('.')?.last() ?: "unknown"
 
-    var attrs = listOf(
-        Pair("idEntry", idEntry ?: emptyString),
-        Pair("idPackage", idPackage ?: emptyString),
-        Pair("idType", idType ?: emptyString),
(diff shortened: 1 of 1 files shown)


app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
```diff
@@ -15,21 +17,21 @@ private fun AssistStructure.ViewNode.dumpNode(sb: StringBuilder = StringBuilder(
...
+        Pair("autofillHints", autofillHints?.joinToString(", ") ?: emptyString),
+        Pair("autofillOptions", autofillOptions?.joinToString(", ") ?: emptyString)
     )
 
     htmlInfo?.attributes?.let { attributes ->
         attrs += attributes.map { Pair(it.first, it.second) }
     }
 
-    attrs = attrs.filter { it.second != "" && it.second != "null" }
+    attrs = attrs.filter { it.second != emptyString && it.second != "null" }
```
jhugman: Given that we're filtering here for the placeholders `emptyString` can we get away with having `String?` here instead of `String`?

Also, if you're touching this code, please could you consider replacing `Pair(x, y)` with `x to y`.

app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
```diff
@@ -15,21 +17,21 @@ private fun AssistStructure.ViewNode.dumpNode(sb: StringBuilder = StringBuilder(
     val name = htmlInfo?.tag ?: className?.split('.')?.last() ?: "unknown"
 
     var attrs = listOf(
-        Pair("idEntry", idEntry ?: ""),
-        Pair("idPackage", idPackage ?: ""),
-        Pair("idType", idType ?: ""),
-        Pair("webDomain", webDomain ?: ""),
-        Pair("hint", hint ?: ""),
-        Pair("autofillValue", autofillValue?.textValue ?: ""),
```
jhugman: Is this the line that is throwing the exception?

Looking at the implementation of `getTextValue()` it looks like we should be testing `autofillValue?.isText()` before calling `.textValue`.

Perhaps:
```kotlin
"autofillValue" to autofillValue?.let { if (it.isText()) it.textValue } ?: emptyString
```

eliserichards: That is what I'm doing in 
```
Pair("autofillValue", autofillValue?.isText ?: autofillValue?.textValue ?: emptyString)
```

I can change the syntax to use `let` instead of `?:`s if that's preferred

app/src/main/java/mozilla/lockbox/extensions/ViewNode+.kt
```diff
@@ -14,22 +16,37 @@ fun AssistStructure.ViewNode.dump(): String {
...
+            }
     )
 
     htmlInfo?.attributes?.let { attributes ->
-        attrs += attributes.map { Pair(it.first, it.second) }
+        attrs += attributes.map { it.first to it.second }
     }
 
-    attrs = attrs.filter { it.second != "" && it.second != "null" }
+    attrs = attrs.filter { it.first != emptyString && it.second != "null" }
```
jhugman: I think there's a misunderstanding of what this code is doing.

In your comment: 

```kotlin
autofillValue?.isText ?: autofillValue?.textValue ?: emptyString
```

`autofillValue?.isText ?` is a `Boolean?`. The `?:` elvis operator is a fallthrough if the left hand side is a `null`.

So, if `autofillValue` doesn't exist, then we try `autofillValue?.textValue`, which is of type `String?`, but we know that it's going to be `null`, so we get `emptyString`.

The interesting bit is what type this is going to be. The compiler thinks it's going to be a `Boolean || String` optional. Since kotlin doesn't have algebraic types, then the only thing it can conclude is that it's `Any?`.

The code you've committed:

```kotlin
            if (autofillValue != null && autofillValue?.isText!!) {
                autofillValue!!.textValue.toString()
            } else {
                emptyString
            },
```
works, but requires you force dereference (`!!`). Channeling Matz here, we should treat `!!` as syntactic vinegar: force-dereferencing should be avoided — because of the danger of `NullPointerException` — so they made it ugly to read.

Here's what I suggest:

```kotlin
    val viewAttrs = listOf(
        "idEntry" to idEntry,
        "idPackage" to idPackage,
        "idType" to idType,
        "webDomain" to webDomain,
        "hint" to hint,
        "autofillValue" to autofillValue?.let { if (it.isText) it.textValue.toString() else null },
        "autofillHints" to autofillHints?.joinToString(", "),
        "autofillOptions" to autofillOptions?.joinToString(", ")
    )

    val htmlAttrs = htmlInfo?.attributes?.let { attributes ->
        attributes.map { it.first to it.second }
    } ?: emptyList()

    val attrs = (viewAttrs + htmlAttrs).filter {
        it.second != null
    }
```

The `autofillValue`, `autofillHints` and `autofillOptions` all are nullable. `autofillValue` **must** be checked as to be of text type. If not, we emit `null`.

Then we filter out all the `null` valued pairs.

FWIW, I'm not especially happy we can't use 

```kotlin
val htmlAttrs = htmlInfo?.attributes?.toList() ?: emptyList()
```

but I'm not sure how to persuade kotlin that `Pair<String!, String!>` and `Pair<String!, String!>!` should be treated in the same way.

app/src/test/java/mozilla/lockbox/presenter/AutofillLockedPresenterTest.kt
```diff
@@ -105,6 +106,6 @@ class AutofillLockedPresenterTest : DisposingTest() {
         Mockito.`when`(fingerprintStore.isKeyguardDeviceSecure).thenReturn(true)
 
         settingStore.unlockWithFingerprintStub.onNext(true)
-        dispatcherObserver.assertValue { it is RouteAction.DialogFragment.FingerprintDialog }
+        dispatcherObserver.assertValue { it is RouteAction.DialogFragment.FingerprintDialog }.await(1, TimeUnit.SECONDS)
```
sashei: rather than using an await, can you block on the `register` method of the dispatcher? that's usually more stable.

PR #644, opened 2019-05-01
More reliably clear the clipboard

linuxwolf:
Fixes _TBD_

This patch re-uses the existing broadcast receiver and an alarm to clear the clipboard.  The set alarm is using the most power-friendly configuration; it will fire if the device is running in some form, will not wake it from sleep, should fire at next wakeup, and will not survive a reboot.

## Testing and Review Notes

0. Lockbox running in foreground, with an entry's details displayed
1. tap the copy for either Username or Password
2. send Lockbox to background
3. wait at least 60 seconds
4. attempt to paste

**Expected** nothing (technically the empty string is pasted)

## To Do

- [x] add unit tests


linuxwolf:
The changes to the `lint-baseline.xml` are because the `ClipboardStore` needs the `Context`.  It _should_ be ok since the `ClipboardStore` exists for the entire life of the application anyway ...

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.content.ClipboardManager
+import android.content.Context
+
+typealias ClipboardSupportFactory = (Context) -> ClipboardSupport
+
+open class ClipboardSupport(
+    private val clipboard: ClipboardManager
+) {
+    companion object {
+        val create: ClipboardSupportFactory = { context ->
```
sashei: using `create` instead of `shared` to denote that this is created on-demand and not allocated for the lifecycle of the containing `ClipboardSupport` class

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
```diff
@@ -0,0 +1,36 @@
...
+            ClipboardSupport(context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager)
+        }
+    }
+
+    open fun paste(label: String, value: String) {
+        clipboard.primaryClip = ClipData.newPlainText(label, value)
+    }
+
+    open fun clear(dirty: String, clean: String = "") {
```
sashei: do we need a `clean` parameter?

app/src/main/java/mozilla/lockbox/LockboxBroadcastReceiver.kt
```diff
@@ -0,0 +1,46 @@
...
+                .apply()
+        }
+    }
+
+    private fun clearClipboard(context: Context?, intent: Intent?) {
+        context?.let {
+            val support = createClipboardSupport(it)
+            val dirty = intent?.getStringExtra(Constant.Key.clipboardDirtyExtra)
+
+            support.clear(dirty ?: "", "")
```
sashei: Let's use an `emptyString` variable here rather than the actual empty string.

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
sashei: why would you use this strategy instead of a convenience initializer that takes a `Context`?

app/src/main/java/mozilla/lockbox/store/ClipboardStore.kt
```diff
@@ -46,25 +54,22 @@ open class ClipboardStore(
...
-            replaceDirty(dirty, clean)
-        }, delay)
-    }
-
-    fun replaceDirty(dirty: String, clean: String = "") {
-        val clipData = clipboardManager.primaryClip?.getItemAt(0)
-        if (clipData?.text == dirty) {
-            clipboardManager.primaryClip = ClipData.newPlainText("", clean)
-        }
+    private fun timedReplaceDirty(dirty: String, delay: Long = defaultClipboardTimeout) {
```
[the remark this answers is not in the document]
linuxwolf: It does.  I felt it better to have one place a leak-potential action is happening than two.

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
linuxwolf: TBH, I was trying out a few things to get to the point of working, and didn't go back to revisit this.  Looking it over, I think just having the initializer take a `Context` handles all the cases.

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
linuxwolf: We don't.  This is a leftover.

app/src/main/java/mozilla/lockbox/support/ClipboardSupport.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package mozilla.lockbox.support
+
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
+
+typealias ClipboardSupportFactory = (Context) -> ClipboardSupport
+
+open class ClipboardSupport(
```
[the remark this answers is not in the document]
linuxwolf: It felt like the lightest-touch method to make the receiver more testable

PR #653, opened 2019-05-06
610: Rebrand Lockbox to new Lockwise brand, name, colors, etc.

### LINKED ISSUE #610: Rebrand to Lockwise
[android icon assets.zip](https://github.com/mozilla-lockbox/lockbox-android/files/3104221/android.icon.assets.zip)

Zeplin: https://zpl.io/Vk5l0AZ
InVision: https://mozilla.invisionapp.com/share/3SN35UHWJDX
Rollout plan: https://docs.google.com/document/d/1U-Rqu4WZAYrLbwPeTM8F32Zwg67ymEq9CSRjhQ2w2Ec/edit

devinreams:
Fixes #610


Pike:
I see that you're changing the entry IDs to account, that's one of the questions I had on the strings.xml.

I also see that `learn_to_edit_entry` is still talking about entry. And there are a few comments about "detailed item entry view". I wonder if that's good or bad?

Pike:
Keyboard was eager to submit. Overall, the `strings.xml` looks good, I'm trying to add comments on the details as things progress here.

devinreams:
Thanks for the extra eyes @Pike!

I'm taking another pass at finding: entry/entries, login/logins, Lockbox/lockbox, "your". Just pushed the latest here and I think we're nearly stable. Do you happen to spot anything else?

app/src/main/res/values/strings.xml
```diff
@@ -20,7 +20,7 @@
     <!-- This is used as the label on a button on a first-run welcome screen to sign in and start using the application -->
     <string name="welcome_start_btn">Get Started</string>
     <!-- This provides context near a button on the first-run welcome screen to describe application requirements. Firefox Account is to be translated consistently to match use in other Firefox products. %1$s will be replaced with application name. -->
-    <string name="welcome_instructions">To use %1$s, you’ll need a\nFirefox Account with saved logins.</string>
+    <string name="welcome_instructions">To use %1$s, you’ll need a Firefox\nAccount with saved logins.</string>
```
Pike: This indicates that we want to have on-device testing for this string. Automated screenshot testing goes a long way on problems like this. CC @Delphine to point at other engineers that have done so successfully.

app/src/main/res/values/strings.xml
```diff
@@ -198,24 +198,24 @@
...
+    <string name="no_logins_found">No Logins Found</string>
+    <!-- This explains the steps the user should take, if no saved logins are found: sign in and sync in Firefox. %1$s will be replaced with application name -->
+    <string name="no_logins_description">%1$s lets you access passwords you’ve already saved to Firefox. To
+        view your logins here, you’ll need to sign in and sync in Firefox.
     </string>
     <!-- This is the title for the biometric unlock onboarding screen. %1$s will be replaced with application name -->
     <string name="onboarding_unlock_title">Unlock %1$s with your fingerprint.</string>
     <!-- This is the onboarding description for enabling biometric unlock with the fingerprint sensor. %1$s will be replaced with application name -->
-    <string name="onboarding_unlock_description">Give %1$s permission to unlock this app with the fingerprint sensor.</string>
+    <string name="onboarding_unlock_description">Give %1$s permission to unlock this\napp with the fingerprint sensor.</string>
```
Pike: Is there a way to not require a manual line break here?

If not, this is another case where being able to verify with actual screenshot tests goes a long way.

app/src/main/res/values/strings.xml
```diff
@@ -238,7 +238,7 @@
     <!-- This is the title of the device security dialog box. -->
     <string name="secure_your_device">Secure your device.</string>
```
Pike: Should this be title-case? Honest question.

app/src/main/res/values/strings.xml
```diff
@@ -249,8 +249,6 @@
 
     <!-- This is the prompt when autofill is triggered and the application is locked. %1$s will be replaced with application name -->
     <string name="autofill_authenticate_cta">Unlock %1$s</string>
-    <!-- This is the error message when autofill is triggered but no domain is detected. This is unlikely. -->
-    <string name="autofill_error_no_hostname">Unexpected hostname format</string>
     <!-- This is the prompt when autofill is triggered, application is unlocked, and no passwords are detected. %1$s will be replaced with application name. -->
     <string name="autofill_search_cta">Search %1$s</string>
     <!-- This is the error message when autofill is triggered and the application is in an error state. %1$s will be replaced with application name. %s refers to the localized error message that will be provided by the system. -->
```
Pike: `%s refers to the localized error message` -> `%2$s`

app/src/main/res/values/strings.xml
```diff
@@ -249,8 +249,6 @@
 
     <!-- This is the prompt when autofill is triggered and the application is locked. %1$s will be replaced with application name -->
     <string name="autofill_authenticate_cta">Unlock %1$s</string>
-    <!-- This is the error message when autofill is triggered but no domain is detected. This is unlikely. -->
-    <string name="autofill_error_no_hostname">Unexpected hostname format</string>
     <!-- This is the prompt when autofill is triggered, application is unlocked, and no passwords are detected. %1$s will be replaced with application name. -->
     <string name="autofill_search_cta">Search %1$s</string>
```
Pike: Is there a better way to describe what "Search Lockwise" actually does? This sounds like something that could mean "search for" or "search in" or "search with".

app/src/main/res/values/strings.xml
Delphine: @npark-mozilla has been helping us for a long time with screenshots on Android. And I don't know anyone on Lockbox side that has been involved with this, but maybe No-Jun can share some knowledge?

npark-mozilla: @Delphine so in other projects, we have been using fastlane tools and espresso UI test script that goes through the screens and generate l10n screenshots.  It has it's quirks, but it does scale well if the number of screenshots increase (also easy to switch locales).  Only caveat is that it does not work with kotlin, so the espresso test must be written in Java.

app/src/main/res/values/strings.xml
devinreams: The design is intentionally a sentence and sentence case

Screen Shot 2019-05-08 at 12 51 54 PM

I'll let @changecourse chime in and _course_ correct us if we should instead do a heading/title case instead... but will assume not for now and mark "resolved"

app/src/main/res/values/strings.xml
devinreams: Good call out here, thanks.

After discussing with @sashei and @eliserichards we will not use a newline and instead explore  some layout constraints to get closer to the intended look (and line wrapping).

app/src/main/res/values/strings.xml
devinreams: Good call out here, thanks.

After discussing with @sashei and @eliserichards we will not use a newline and instead explore  some layout constraints to get closer to the intended look (and line wrapping).

PR #694, opened 2019-05-22
649: disconnect account security bugs

### LINKED ISSUE #649: Bug #1548361, #1549750 - Disconnect account
https://bugzilla.mozilla.org/show_bug.cgi?id=1548361 - webview cache deletion
https://bugzilla.mozilla.org/show_bug.cgi?id=1549750 - passwords/database (locally, delete encryption key for db)

eliserichards:
Fixes #649

- [x] delete local storage logs
- [x] delete webview caches
- [x] remove encryption key when disconnecting account
 
## Testing and Review Notes
**Note:** `getDir("..")` was not working to find the whole correct path to the logs, so I used this instead:
```
logDirectory = context.getDir("webview", Context.MODE_PRIVATE)
...
val leveldbDir = logDirectory.listFiles()
                    .filter { file ->
                        file.name.startsWith("Local")
                    }[0]
                    .listFiles()
                    .filter { file ->
                        file.name.startsWith("leveldb")
                    }[0]
                    .listFiles()
```

Steps to reproduce (from the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1548361))

1. Sign in to Lockbox with a valid Firefox user account
2. Disconnect from the application by using Account > “Disconnect Firefox Lockbox”;
3. Confirm the decision on the prompted dialog;
4. Inspect the following paths, reviewing each file content with e.g. xxd:
   * /data/data/mozilla.lockbox/app_webview/Local Storage/leveldb/[0-9]*.log
   * /data/data/mozilla.lockbox/cache/org.chromium.android_webview/*

## To Do

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -9,11 +9,9 @@ package mozilla.lockbox.store
 import android.content.Context
 import android.net.Uri
 import android.os.Looper
-import android.text.format.DateUtils
 import android.webkit.CookieManager
 import android.webkit.WebStorage
 import android.webkit.WebView
-import android.webkit.WebViewClient
 import io.reactivex.Observable
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.addTo
@@ -44,8 +42,6 @@ import mozilla.lockbox.support.Optional
 import mozilla.lockbox.support.SecurePreferences
 import mozilla.lockbox.support.asOptional
 import java.io.File
-import java.lang.Exception
-import java.util.Date
 import java.util.concurrent.TimeUnit
 import kotlin.coroutines.CoroutineContext
 
@@ -69,7 +65,6 @@ open class AccountStore(
             )
         }
 
-    private lateinit var context: Context
     private val coroutineContext: CoroutineContext
         get() = Dispatchers.Default + exceptionHandler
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -224,6 +232,33 @@ open class AccountStore(
 
         this.securePreferences.remove(Constant.Key.firefoxAccount)
         this.generateNewFirefoxAccount()
+        webView.clearCache(true)
+        clearLogs()
+    }
+
+    private fun clearLogs() {
```
sashei: I'm not sure why these methods are included. We don't write anything sensitive to the logs, so clearing them out when disconnecting seems like extra overhead. Is this related the the webview cache?

sashei: oops, just re-read that bug. this looks good to me.

eliserichards:
As part of bug [#1548361](https://bugzilla.mozilla.org/show_bug.cgi?id=1548361), the suggestions to mitigate this would be to 1. clear webview caches and 2. clear LocalStorage’s LevelDB log files on account disconnect.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -224,6 +232,41 @@ open class AccountStore(
...
+                    .filter { file ->
+                        file.name.startsWith("Local")
+                    }[0]
+                    .listFiles()
+                    .filter { file ->
+                        file.name.startsWith("leveldb")
+                    }[0]
+                    .listFiles()
+
+                for (logs in leveldbDir) {
```
sashei: nit: putting an `s` after something that's really a singular piece of the `for`-loop. maybe `for file in...` or `for logFile in...`.

PR #695, opened 2019-05-23
Android Studio 3.4

hannesa2:
It's simply an update to most recent stable Android Studio


hannesa2:
> Hi @hannesa2! This is great! Can you add a comment here that explains the `@ignore("...")` statements that you have added? (For example, in
> 
> ```
> @Ignore("619-ui-tests-bitrise (#620)")
> ```
> 
> is `619-ui-tests-bitrise` the branch name?

@eliserichards I run into lint errors during this PR and one source where this `@Ignore`. Here lint asked for a ignore reason. So either I could add an additional `@SuppressLint("IgnoreWithoutReason")` or add a reason. So I decide to add as reason the last commit message (just the first line)

eg:
image

hannesa2:
Btw, you use a `lint-baseline.xml` 
In my point of view, I would remove this complete. With this

PR #709, opened 2019-05-24
673: Added a small delay to navigate to the confirmation screen

### LINKED ISSUE #673: Onboarding confirmation screen delay
Currently, when someone is on the Autofill onboarding screen, and then goes to set their autofill provider via the "Go to Settings" button, the Confirmation "You're all Set" overlay is displayed before the app switches them to the settings screen... 

Ideally, we would stay on the Autofill screen until they choose Lockwise as an autofill provider, and only navigate to the "You're all set" confirmation once they've been successfully routed back to Lockwise. 

Alternatively, we could set a delay on the button for "Go to Settings" so that the confirmation screen is triggered after they've already been switched to the settings screen for autofill, though this may cause unforeseen issues... not sure.

### LINKED ISSUE #697: Lockwise logo is flashing when resuming Lockwise while being in "You're all set!" screen
### Steps to reproduce
1. Launch lockwise.
2. Login with valid credentials.
3. When reaching **`You're all set!`** put lockwise in the background.
4. Resume lockwise.
### Expected behavior
Lockwise is resumed and no display issues are present.
### Actual behavior
When resuming lockwise, the lockwise logo is displayed right in the middle of the **`You're all set!`** screen.
### Device and build information
* **`Google Pixel 3(Android 9)`**
*  **`Lockbox version: 1.1.0 (3951)`**
### Notes 
This issue can be reproduced on any device, for more details please see the attachments: [video](https://drive.google.com/file/d/1YF9Q6FsuSjphfrjqwDs93r1fE-1cKfVW/view?usp=sharing).

faogustavo:
Fixes #673
Fixes #697

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/AutofillOnboardingPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/AutofillOnboardingPresenter.kt
@@ -41,14 +41,15 @@ class AutofillOnboardingPresenter(
         view.onGoToSettingsClick
             .subscribe {
                 dispatcher.dispatch(RouteAction.SystemSetting(SettingIntent.Autofill))
-                triggerNextOnboarding()
+                triggerNextOnboarding(delayed = true)
             }
             .addTo(compositeDisposable)
     }
 
-    private fun triggerNextOnboarding() {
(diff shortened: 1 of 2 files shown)


faogustavo:
I've just realized that if you click on the other button, you will also have the delay. I'll fix it soon.

PR #710, opened 2019-05-25
646: Adjusted to avoid duplicate the top fragment

### LINKED ISSUE #646: Opening a website from entry details requires two "back" taps after returning to Lockbox
steps to reproduce:
- open a website from an entry detail screen
- return to Lockbox (device back button, app switcher)
- tap back once on the entry detail

**expected**
the app navigates back to the item list

**actual**
the app stays on the entry details and requires a second "back" tap to return to the item list

faogustavo:
Fixes #646

## Testing and Review Notes
The problem was at the `AppRoutePresenter`. Whenever the app goes to background and returns, we were subscribing to the `routeStore`. This will dispatch a navigation action and duplicated the top fragment.


devinreams:
Thank you for your continued contributions and help @faogustavo! 😍

PR #720, opened 2019-05-29
Remove unnecessary roboelectric testrunners

eliserichards:
Fixes N/A

## Testing and Review Notes
General cleanup that will speed up our tests a little bit. Not related to any issue.

Average `android-unit-test` times from last 5 master builds: 146.2 s
Average from 5 builds of this branch: 127 s

## To Do


app/src/test/java/mozilla/lockbox/presenter/FilterPresenterTest.kt
```diff
@@ -79,7 +79,7 @@ class FilterPresenterTest {
 
     val view = FakeFilterView()
     val dispatcher = Dispatcher()
```
sashei: should these other properties be private also?

PR #732, opened 2019-06-03
648: Remove registered device from AS device list

### LINKED ISSUE #648: Bug #1548367 - sessionToken on disconnect
https://bugzilla.mozilla.org/show_bug.cgi?id=1548367

Relies on invalidate_session_token method in app-services.

depends on https://github.com/mozilla/application-services/issues/1139

eliserichards:
Fixes #648

## Testing and Review Notes
This implementation parses the Fx Account into the application services Fx Account, which contains the list of devices that are associated with that account. From there, we can find the current device and remove it from the list. 

## To Do

- [ ] ~add unit tests~

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -20,7 +20,6 @@ import io.reactivex.subjects.Subject
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.rx2.asMaybe
 import kotlinx.coroutines.rx2.asSingle
 import mozilla.appservices.fxaclient.FxaException
@@ -243,8 +242,14 @@ open class AccountStore(
     }
 
     private fun removeDeviceFromFxA() {
-        runBlocking {
-            fxa?.deviceConstellation()?.destroyCurrentDeviceAsync()?.await()
+        if (fxa != null) {
+            fxa!!.deviceConstellation()
+                .destroyCurrentDeviceAsync()
+                .asSingle(coroutineContext)
+                .subscribe()
+                .addTo(compositeDisposable)
+        } else {
+            log.info("FxA is null. No devices to disconnect.")
         }
     }
 
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -243,17 +243,17 @@ open class AccountStore(
     }
 
     private fun removeDeviceFromFxA() {
-        if(fxa != null) {
-            val account = mozilla.appservices.fxaclient.FirefoxAccount.fromJSONString(fxa!!.toJSONString())
-            val registeredDevices = account.getDevices()
+        val account = mozilla.appservices.fxaclient.FirefoxAccount
+            .fromJSONString(
+                fxa?.toJSONString() ?: ""
+            )
+        
+        val registeredDevices = account.getDevices()
 
-            for (device in registeredDevices) {
-                if(device.isCurrentDevice) {
-                    account.destroyDevice(device.id)
-                }
+        for (device in registeredDevices) {
+            if(device.isCurrentDevice) {
+                account.destroyDevice(device.id)
             }
-        } else {
-            log.info("FxA is null. ")
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -235,8 +237,22 @@ open class AccountStore(
 
         webView.clearCache(true)
         clearLogs()
+        removeDeviceFromFxA()
```
sashei: Why did we lose the `DataStoreAction.Reset`?

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -235,8 +237,22 @@ open class AccountStore(
...
+    }
+
+    private fun removeDeviceFromFxA() {
+        if(fxa != null) {
+            val account = mozilla.appservices.fxaclient.FirefoxAccount.fromJSONString(fxa!!.toJSONString())
```
sashei: Whoa, a lot going on here! We convert the `fxa` to JSON and then convert it back!

Why would we do that as opposed to just using the current FxA instance? I can see why things would get tripped up here because the calls to remove the previous stored instance and generate a new one occur _before_ we go through to remove the device, but that should be ameliorated by moving those calls after the `removeDeviceFromFxA` call (`this.securePreferences.remove(Constant.Key.firefoxAccount)` and `this.generateNewFirefoxAccount()` specifically). 

Also, checking to see if something is null usually means that we can use the `let` syntax. So this could read:
```kotlin
fxa?.let {
   val registeredDevices = it.getDevices()
    for (device in registeredDevices) {
        if(device.isCurrentDevice) {
             account.destroyDevice(device.id)
         }
    }
}
```
or even
```kotlin
val registeredDevices = fxa?.getDevices ?: []
for (device in registeredDevices) {
   if(device.isCurrentDevice) {
      account.destroyDevice(device.id)
   }
}
```

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
eliserichards: It was redundant here (per our discussion when I added the `clearLogs()`) but I forgot to remove it in that PR before it was merged.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
eliserichards: We can't access the `getDevices` method from an a-c FxA account. It can only be retrieved from an a-s FxA client. The companion obj in the FxA client says to do the conversion from a-c using the json method: 
```
companion object {
        /**
         * Restores the account's authentication state from a JSON string produced by
         * [FirefoxAccount.toJSONString].
         *
         * This does not make network requests, and can be used on the main thread.
         *
         * @return [FirefoxAccount] representing the authentication state
         */
        fun fromJSONString(json: String, persistCallback: PersistCallback? = null): FirefoxAccount {
            return FirefoxAccount(rustCall { e ->
                LibFxAFFI.INSTANCE.fxa_from_json(json, e)
            }, persistCallback)
        }
    }
```

eliserichards: https://github.com/mozilla/application-services/blob/master/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FirefoxAccount.kt

eliserichards: (I like the `let` syntax and I am going to add that)

sashei: OK, then I think we should request that the android-components version of `FirefoxAccount` reflects the `application-services` implementation. They should be the same thing, or we should be using the `application-services` version everywhere.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -237,8 +240,12 @@ open class AccountStore(
 
         webView.clearCache(true)
         clearLogs()
+    }
 
-        dispatcher.dispatch(DataStoreAction.Reset)
+    private fun removeDeviceFromFxA() {
+        runBlocking {
```
eliserichards: Not sure if `run {}` or `runBlocking {}` would be a suitable coroutine to use here. Maybe you have thoughts @sashei

sashei: whoa, we should never block the thread in regular code! let's stick with `run` for now... or even better, create an Observable to wrap this code so that we can interact with it in the usual way :)

eliserichards: Let's pair on that. I'm not sure how to create an Observable for a `Deferred` type.

PR #746, opened 2019-06-06
685: Dispatch action to toggle password visibility to false on background

### LINKED ISSUE #685: Bug #1549765 - toggle password visibility in background
https://bugzilla.mozilla.org/show_bug.cgi?id=1549765

This ticket will track toggling the visibility of passwords whenever the app is backgrounded. For now we only care about passwords being visible, and toggle the password to hidden every time we go to the background. Further work will be done as part of #689.

eliserichards:
Fixes #685

## Testing and Review Notes
This toggles the visibility of the password to "off" when the `ItemDetailFragment` is disposed.

## Screenshots or Videos
image

## To Do

- [x] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
@@ -18,7 +18,6 @@ import mozilla.lockbox.action.ClipboardAction
 import mozilla.lockbox.action.DataStoreAction
 import mozilla.lockbox.action.ItemDetailAction
 import mozilla.lockbox.action.RouteAction
-import mozilla.lockbox.extensions.debug
 import mozilla.lockbox.extensions.filterNotNull
 import mozilla.lockbox.extensions.toDetailViewModel
 import mozilla.lockbox.flux.Dispatcher
@@ -113,7 +112,6 @@ class ItemDetailPresenter(
             .addTo(compositeDisposable)
 
         itemDetailStore.isPasswordVisible
-            .debug("SUBSCRIBE TO STORE")
             .subscribe { view.isPasswordVisible = it }
             .addTo(compositeDisposable)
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -102,6 +113,7 @@ class ItemDetailPresenter(
             .addTo(compositeDisposable)
 
         itemDetailStore.isPasswordVisible
+            .debug("SUBSCRIBE TO STORE")
```
sashei: ⚠️

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
```diff
@@ -69,6 +69,16 @@ class ItemDetailFragment : BackableFragment(), ItemDetailView {
             updatePasswordVisibility(value)
         }
 
+    override fun onPause() {
```
sashei: can you add these calls to the base `Fragment` class? or ensure that they don't already exist there?

eliserichards: yes! they didn't exist there before. Is it always best to put those overrides in the `Fragment`?

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
eliserichards: 😱

app/src/main/java/mozilla/lockbox/view/ItemDetailFragment.kt
sashei: I think you're fine leaving the `onDestroyView` instead of `onDestroy` in there... and yeah, the idea with the base `Fragment` class is that individual `Fragment`s don't need to call any additional lifecycle actions (remembering to call `onViewReady` every time etc), you can just implement them in the `Presenter` as needed.

Issue #754, opened 2019-06-12
Syncing on foreground timeout


eliserichards:
Reproducible on the latest release build & latest debug build of master.

Steps to reproduce: 

1. Install app and sign in - **set autofill provider** to Lockwise
2. Select an item and show password
3. Background the app from the item list
4. Foreground the app - see "Syncing has timed out" toast
5. Select an item - instead of routing to the item, see that sync is kicked off again from item list and times out 
6. See the previous logins list loaded after the timeout, even though sync failed

Step 5 does not happen when the autofill provider is not set to Lockwise. 

<img width="350" alt="image" src="https://user-images.githubusercontent.com/43795363/59383784-7d728e80-8d15-11e9-8f38-8551f93eb41a.gif">

eliserichards:
Fixed by #756.

PR #756, opened 2019-06-12
implement sync interval every 24 hours

### LINKED ISSUE #735: App syncs every time it's brought to the foreground
On my phone, Lockwise syncs whenever the app is brought to the foreground, which makes it very hard to use (especially on a slow connection). Right now it even syncs when the screen turns off and on.

I'm not totally sure if this is the intended behavior or if I hit a bug but I'm happy to add whatever debugging information you might need.

Phone: Samsung S10e
Version: Latest from playstore (6/4)

sashei:
Fixes #735 
Connected to #754

## Testing and Review Notes

Currently, manual sync resets the 24 hour timer, so if you would like to see an automatic sync, don't sync manually!

## To Do

- [x] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
+++ b/app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
@@ -17,24 +17,37 @@ import mozilla.lockbox.store.ContextStore
 import mozilla.lockbox.store.SettingStore
 
 @ExperimentalCoroutinesApi
-open class AutoLockSupport(
+open class TimingSupport(
     val settingStore: SettingStore = SettingStore.shared
 ) : ContextStore {
 
     companion object {
-        val shared by lazy { AutoLockSupport() }
+        val shared by lazy { TimingSupport() }
     }
 
     private val compositeDisposable = CompositeDisposable()
     private lateinit var preferences: SharedPreferences
 
-    var lockingSupport: LockingSupport = SystemLockingSupport()
+    var systemTimingSupport: SystemTimingSupport = SystemSystemTimingSupport()
 
     open val shouldLock: Boolean
         get() = lockCurrentlyRequired()
 
+    open val shouldSync: Boolean
+        get() = syncCurrentlyRequired()
+
     override fun injectContext(context: Context) {
         preferences = PreferenceManager.getDefaultSharedPreferences(context)
+    }
+
+    private fun syncCurrentlyRequired(): Boolean {
+        val syncTimerDate = preferences.getLong(Constant.Key.syncTimerDate, Long.MAX_VALUE)
+        val currentSystemTime = systemTimingSupport.systemTimeElapsed
+
+        // this will be triggered when the system time is very low due to device restarts
+        val currentTimeTooEarly = (syncTimerDate - currentSystemTime) > Constant.Common.twentyFourHours
+
+        return syncTimerDate < currentSystemTime || currentTimeTooEarly
     }
 
     open fun storeNextAutoLockTime() {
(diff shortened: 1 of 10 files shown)


app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
```diff
@@ -17,33 +17,50 @@ import mozilla.lockbox.store.ContextStore
...
     }
 
+    private fun syncCurrentlyRequired(): Boolean {
+        val syncTimerDate = preferences.getLong(Constant.Key.syncTimerDate, Long.MAX_VALUE)
+        val currentSystemTime = systemTimingSupport.systemTimeElapsed
+
+        // this will be triggered when the system time is very low due to device restarts
+        val currentTimeTooEarly = (syncTimerDate - currentSystemTime) > Constant.Common.twentyFourHours
+
+        return syncTimerDate < currentSystemTime || currentTimeTooEarly
```
jhugman: Not sure I understand the name and comment `currentTimeTooEarly`.

```kotlin
syncTimerDate > currentSystemTime + twentyFourHours
```

Oh, so if `syncTimerDate` (which should never be less than twenty four hours after the current time) is 24h after the current time, then we've had something weird happen, e.g. a reboot.

So, suggest: renaming `currentTimeTooEarly` to `followingDeviceReboot`.

```kotlin
val followingDeviceReboot = syncTimerDate > currentSystemTime + twentyFourHours
```

jhugman: Incidentally, I'd put 24 hours as a property.

app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
```diff
@@ -17,33 +17,50 @@ import mozilla.lockbox.store.ContextStore
...
+
     open fun storeNextAutoLockTime() {
         settingStore.autoLockTime
             .take(1)
             .subscribe(this::updateNextLockTime)
             .addTo(compositeDisposable)
     }
 
+    open fun storeNextSyncTime() {
+        storeSyncTimerDate(systemTimingSupport.systemTimeElapsed + Constant.Common.twentyFourHours)
```
jhugman: Set `Constant.Common.twentyFourHours` as a property of `TimingSupport`.

e.g. 

```kotlin
private val syncInterval = Constant.Common.twentyFourHours
```

app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
```diff
@@ -56,12 +73,12 @@ open class AutoLockSupport(
...
         } else {
-            storeAutoLockTimerDate(lockingSupport.systemTimeElapsed + autoLockTime.ms)
+            storeAutoLockTimerDate(systemTimingSupport.systemTimeElapsed + autoLockTime.ms)
         }
     }
 
     private fun lockCurrentlyRequired(): Boolean {
-        val currentSystemTime = lockingSupport.systemTimeElapsed
+        val currentSystemTime = systemTimingSupport.systemTimeElapsed
         return if (currentSystemTime <= Constant.Common.sixtySeconds) {
```
jhugman: I'm not sure I understand why `Constant.Common.sixtySeconds` is not in a property — isn't `autoLockInterval` settable via `SharedPreferences`?

app/src/main/java/mozilla/lockbox/support/TimingSupport.kt
sashei: I stuck it in the constants with the expectation that it will eventually become a setting that users can configure per #757 .

PR #872, opened 2019-08-14
Update Readme with l10n process

Delphine:
This is something we've been adding to all Readmes of mobile repos to direct people to Pontoon for localization (vs GitHub repos, where work isn't supposed to happen).


eliserichards:
Looks great @Delphine! I'm going to get this merged right now.

PR #853, opened 2019-07-27
WIP Automatic FxA sign-in support

grigoryk:
This is a functional demo (needs UI/UX polish) of FxA auto-sign functionality which we just landed in https://github.com/mozilla-mobile/android-components/pull/3905

This allows Lockwise to automatically sign-in into an account that's present on Firefox for Android. Lockwise is already whitelisted for FxA auth access on released Firefox for Android.

Sometime soon, Fenix will be able to act as an authentication state provider as well. 

~Before landing this make sure this landed: https://github.com/mozilla-mobile/android-components/pull/3806 (we need API fixes in a-s 0.35.4)~

PR #879, opened 2019-08-16
Auto-signin with mozilla AuthProvider

jhugman:
This PR continues the PR #853.

~This is not ready to land, because it needs:~

~* [ ] UX review for screen design~
~* [ ] Animation, wait states~
~* [ ] Possibility of selecting from multiple accounts.~
~* [ ] Syncing after new signin.~

Filed #894 to help land this (and avoid it bit rotting) behind a feature flag.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
@@ -9,7 +9,7 @@ package mozilla.lockbox.presenter
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.addTo
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import mozilla.components.service.fxa.sharing.AccountSharing
+import mozilla.components.service.fxa.sharing.ShareableAccount
 import mozilla.lockbox.action.AccountAction
 import mozilla.lockbox.action.AppWebPageAction
 import mozilla.lockbox.action.DialogAction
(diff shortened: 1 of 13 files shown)


jhugman:
Detecting email from a Fennec auth provider.

grigoryk:
> Possibility of selecting from multiple accounts.

I wouldn't worry about that too much for a first stab at this. We're just using the first provided account in Fenix. Underlying a-c API will give them in an order of our internal preference - so release first, etc. Once Fenix becomes auth provider, it'll take precedence over fennec.

grigoryk:
> Syncing after new signin.
Once lockwise switches to use FxaAccountManager, this sort of thing will happen automatically. But either way, this should be straightforward.

app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
```diff
@@ -8,31 +8,47 @@ package mozilla.lockbox.presenter
...
 ) : Presenter() {
+
     override fun onViewReady() {
-        view.getStartedClicks
-            .map {
-                if (fingerprintStore.isKeyguardDeviceSecure)
-                    RouteAction.Login
-                else {
-                    DialogAction.OnboardingSecurityDialog
+        accountStore.shareableAccount()?.let { account ->
```
eliserichards: Not sure if I like the `.map{}.subscribe{}` inside of the `.let{}` here

app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
```diff
@@ -43,4 +59,18 @@ class WelcomePresenter(
             }
             .addTo(compositeDisposable)
     }
-}
\ No newline at end of file
+
+    private fun routeToUseExistingAccount(account: ShareableAccount) =
```
eliserichards: Would prefer `routeToExistingAccount` here

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -243,8 +266,7 @@ open class AccountStore(
 
     private fun removeDeviceFromFxA() {
         if (fxa != null) {
-            fxa!!.deviceConstellation()
-                .destroyCurrentDeviceAsync()
+            fxa!!.disconnectAsync()
```
eliserichards: 👏

app/src/main/java/mozilla/lockbox/view/WelcomeFragment.kt
```diff
@@ -28,12 +32,35 @@ class WelcomeFragment : Fragment(), WelcomeView {
         val appLabel = getString(R.string.app_label)
         view.textViewInstructions.text = getString(R.string.welcome_instructions, appLabel)
         view.lockwiseIcon.contentDescription = getString(R.string.app_logo, appLabel)
+
         return view
     }
 
-    override val getStartedClicks: Observable<Unit>
+    override fun showExistingAccount(email: String) {
+        view?.apply {
```
eliserichards: Is the view ever going to be null here?

build.gradle
```diff
@@ -9,7 +9,7 @@
 buildscript {
     ext.android_support_version = '28.0.0'
     ext.kotlin_version = '1.3.0'
-    ext.android_components_version = '7.0.0'
+    ext.android_components_version = '9.0.0-SNAPSHOT'
```
eliserichards: Obv change this to `9.0.0` when it's released

build.gradle
```diff
@@ -25,6 +25,7 @@ buildscript {
     ext.constraintlayout_version = '2.0.0-alpha5'
 
     repositories {
+        mavenLocal()
```
eliserichards: Take this out?

build.gradle
```diff
@@ -45,6 +46,7 @@ buildscript {
 
 allprojects {
     repositories {
+        mavenLocal()
```
eliserichards: Take this out?

app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
jhugman: 👍

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
jhugman: 👍

app/src/main/java/mozilla/lockbox/view/WelcomeFragment.kt
jhugman: No, almost certainly not. I used this because it was the neatest way I could think off to not explicitly deref view repeatedly to get to the buttons. (i.e. `view!!.buttonGetStarted` and `view!!.buttonGetStartedManually`)

build.gradle
jhugman: 👍

build.gradle
jhugman: Yah, this should come out. It's super handy to have in while we're dealing with custom builds of android components.

Now we're moving to use `FeatureFlags`, I'll take it out. Once I pick this work up again (hi, @changecourse ), I'll work on my local machine.

build.gradle
jhugman: Yeppers.

app/src/main/java/mozilla/lockbox/presenter/WelcomePresenter.kt
jhugman: Think of this as an 

```kotlin
val account = accountStore.shareableAccount()
if (account != null) {
   // …
}
```

but safer.

Happy to adopt another suggestion :)

jhugman:
I've put this behind a feature flag as I'd like to land this sooner rather than later, even if it isn't ready.

PR #881, opened 2019-08-19
Import strings from android-l10n

Pike:
State: mozilla-l10n/android-l10n@413e05dba427ef8cbf9e5272ae5cc3b3ccdf91d3

I wonder what's going to become of `sync_timed_out` in the app repo. I'm hesitent to take action on the l10n side, but given how long this has been commented out, is that string just dead?

CC @Delphine to also see the replies.

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -252,5 +252,5 @@
 
     <!-- This is the placeholder text for an empty or null username. -->
     <string name="no_username">(no existe usuario)</string>
-    <!--<string name="sync_timed_out">La sincronización ha expirado</string>-->
+    <string name="sync_timed_out">La sincronización ha expirado</string>
 </resources>
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -252,5 +252,18 @@
 
     <!-- This is the placeholder text for an empty or null username. -->
     <string name="no_username">(Kein Benutzername)</string>
-    <!--<string name="sync_timed_out">Zeitüberschreitung bei Synchronisation</string>-->
+
+    <!-- This is the message displayed when sync has timed out. -->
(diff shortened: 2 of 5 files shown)


eliserichards:
The tentative plan is to remove the sync timeout from our end. Sorry for the ambiguity there! @Pike @Delphine

ddurst:
@Pike -- if we determine that we're removing the timeout, then we won't need the string; that's TBD over the next week or so.

But just to make sure: the vietnamese XML strings are already added to Pontoon and coming in in this patch from there?

Pike:
Yes, they're working on https://pontoon.mozilla.org/vi/android-l10n/tags/ now. Delphine connected the contributor to the team there.

ddurst:
OK, so we'll go ahead and remove the timeout, and those strings will then be gone. The rest of this will exist on the pontoon side.

Pike:
I don't think it's necessary to wait on that roundtrip for this PR.

Also, how do you guys handle the `question` label? Should I have removed that, or are there open questions left, or is that on the person that asked to say "yeah, got a sufficient answer" ?

ddurst:
Eh, no idea. I meant it as a question about whether or not we're removing `timeout` anyway, so I'll remove it. :)

app/src/main/res/values-es/strings.xml
```diff
@@ -252,5 +252,5 @@
 
     <!-- This is the placeholder text for an empty or null username. -->
     <string name="no_username">(no existe usuario)</string>
-    <!--<string name="sync_timed_out">La sincronización ha expirado</string>-->
+    <string name="sync_timed_out">La sincronización ha expirado</string>
```
jhugman: Is this being removed, or uncommented?

Pike: Uncommented. We'll likely want a force-sync on the pontoon side once the string is actually removed. CC @Delphine for that.

eliserichards:
Cleaned up and merged in #893 😄

PR #882, opened 2019-08-20
[WIP] create infrastructure for feature flags

### LINKED ISSUE #871: Create infrastructure for feature flags
### Vision statement / What / Requirements
Put the infrastructure in place to handle features being flagged on/off. This will require the creation of a file to track the constants (i.e. "true" state is enabled, "false" state is disabled).

eliserichards:
Fixes #871 

## Testing and Review Notes
- [ ] create initial infrastructure to handle feature flags
- [ ] add initial flag for the delete feature that's already in `master`
- [ ] add flag handling in tests for delete

## To Do

- [ ] add unit tests


app/src/main/java/mozilla/lockbox/support/Constant.kt
```diff
@@ -26,6 +26,14 @@ object Constant {
         val syncTimeout: Long = 20
     }
 
+    /*
+     *  Used to turn features in the app on and off, where true denotes that the feature is available when
+     *  the app is run and false denotes that the feature is not available when the app is run.
+     */
+    object FeatureToggle {
+        const val delete = false
+    }
```
jhugman: I think I'd break `FeatureToggle` out into its own file. I'd also use longer names like just `delete`.

We're going to put all the logic to decide if we're going to expose a feature in this class, and then just use that class where we implement this feature.

app/src/main/java/mozilla/lockbox/presenter/ItemDetailPresenter.kt
```diff
@@ -61,6 +64,10 @@ class ItemDetailPresenter(
     }
 
     override fun onViewReady() {
+
+        // handle feature toggle
+        if (Constant.FeatureToggle.delete) view.showKebabMenu() else view.hideKebabMenu()
```
jhugman: Nit: use braces and more lines here.

app/src/main/java/mozilla/lockbox/support/Constant.kt
jhugman: We'll want to use calculated properties looking at `BuildConfig` properties here, and most likely setting something up in `build.gradle` to generate build flavors.

PR #884, opened 2019-08-21
867 - Fix obvious NPE sync crash

### LINKED ISSUE #867: KotlinNullPointerException in DataStore.sync
Context from thread in #gv: 

ddurst  5 days ago
1) We have a crash that by far outweighs everything else: kotlin.KotlinNullPointerException in mozilla.lockbox.store.DataStore.sync

ddurst  5 days ago
(and we see the corresponding jump in crashes when we released 1.1.3, which included almost exclusively new telemetry dispatches)

ddurst  5 days ago
2) I'm confused by the notion (which might be wrong) that when lockwise initiates a sync request, we don't really want that to be async -- but... it is, isn't it? Which means we could dispatch a sync_timeout (heck, even an sync_end) and that sync will actually succeed.
...
grisha:biking:  5 days ago
@ddurst so if i'm reading that code right, the timeout is certainly not necessary. It's not clear to me at a glance if a TimeoutException thrown on the scheduler will somehow negatively interact with the running sync - your comments imply that no, sync will continue on. To start, i'd remove that timeout logic entirely. Internally, sync library has its own limits for network timeouts and such, and depending on which version of a-s you're on, will return you with a very detailed SyncReport.

grisha:biking:  5 days ago
When you initiate a sync, you initiate an operation that may take quite some time (in both computer and human terms). We model this in Fenix via observables; you subscribe using a SyncStatusObserver, and are told when sync is running, when it stopped, when it failed, etc.

grisha:biking:  5 days ago
Medium-term plan for lockwise's sync & accounts implementation is to adopt a-c's high-level account & sync APIs - see https://github.com/mozilla-mobile/android-components/tree/master/components/service/firefox-accounts#high-level-apis-recommended-for-most-applications

grisha:biking:  5 days ago
I think the only blocker for that work is missing dataprotect integration for account storage in that high level API.

jhugman:
Fixes #867 

This PR fixes an obvious dereference of a null pointer. 

We have been unable to reproduce the original crash, but we need to fix this soon.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/DataStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/DataStore.kt
@@ -4,6 +4,7 @@
 
 package mozilla.lockbox.store
 
+import androidx.annotation.VisibleForTesting
 import com.jakewharton.rxrelay2.BehaviorRelay
 import com.jakewharton.rxrelay2.ReplayRelay
 import io.reactivex.Observable
@@ -60,7 +61,8 @@ open class DataStore(
 
     internal val compositeDisposable = CompositeDisposable()
     private val stateSubject = ReplayRelay.createWithSize<State>(1)
-    private val syncStateSubject = BehaviorRelay.createDefault<SyncState>(SyncState.NotSyncing)
+    @VisibleForTesting
+    val syncStateSubject: BehaviorRelay<SyncState> = BehaviorRelay.createDefault(SyncState.NotSyncing)
     private val listSubject: BehaviorRelay<List<ServerPassword>> = BehaviorRelay.createDefault(emptyList())
     private val deletedItemSubject = ReplayRelay.create<Consumable<ServerPassword>>()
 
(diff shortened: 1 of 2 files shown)


eliserichards:
The sync support tokens are no longer valid after a short period of time (an hour, maybe?). When talking to the sync team, we should be resetting `syncConfig` every time we sync (assuming that it is no longer valid). From that conversation I would assume that the `SyncUnlockInfo` can _never_ be null. 

Are we assuming that this is a bug on their end? Or possibly in our `resetSupport()`?

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -238,10 +238,15 @@ open class DataStore(
     private fun sync() {
         resetSupport(support)
```
eliserichards: Can we find a better name for this method to make it more clear?

jhugman: This should be addressed in #885

jhugman:
Added tests, and filed followup.

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -235,13 +237,21 @@ open class DataStore(
         }
     }
 
-    private fun sync() {
+    @VisibleForTesting(
```
eliserichards: This is _fantastic_!

app/src/main/java/mozilla/lockbox/store/DataStore.kt
eliserichards: Agreed.

jhugman:
Update this is rolled out to 20% of users over the weekend. No obvious side effects.

Proposal: roll out to 100% of users from today.

PR #889, opened 2019-08-27
791: Remove fake timeout from sync

eliserichards:
Fixes #791

## Testing and Review Notes
The toast notification was removed in #833, but the actual logic was still present. This PR removes all of that (if we want a timeout, it should be implemented on the sync side and then handled by us).

## To Do

- [x] double check the original issue to confirm it is fully satisfied
- [x] add testing notes and screenshots in PR description to help guide reviewers
- [x] remove ~add~ unit tests
- [ ] ~request the "UX" team perform a design review (if/when applicable)~
- [x] make sure CI builds are passing (e.g.: fix lint and other errors)

PR #893, opened 2019-08-30
Import strings from l10n

eliserichards:
Cleaning up lint errors from Pike's branch in PR #881

## To Do


eliserichards:
I'm going to remove them all as part of #889 😄

Issue #791, opened 2019-06-26
Remove Sync timeout


LaurentiuApahideanSV:
## Steps to reproduce
1. Open Lockwise and login with a Mozilla account that has saved logins.
2. Drag down from the top of the screen in order to trigger the sync.

### Expected behavior
The saved logins from the Mozilla account should be displayed.

### Actual behavior
The "Syncing has timed out error" is displayed and the logins are not retrieved.

### Device & build information
* Device: Samsung Galaxy S9 (Android 8.0.0)
* Build version: Lockwise version 1.1.2 (Build 4342)

### Notes
The issue only occurs when using the account laursv52@mailinator.com pass: softvision
Attachments: 
![Syncing has timed out error is displayed constantly](https://user-images.githubusercontent.com/50983519/60181246-76c23c00-982a-11e9-82d7-c3e3beaf7b4e.gif)
[logcat.txt](https://github.com/mozilla-lockwise/lockwise-android/files/3330073/logcat.txt)

### Eng todo
- [ ] add telemetry around timeout

abodea:
Please note that this is the only device that had this issue, on others everything worked correctly.

eliserichards:
This should be fixed with Sasha's PR #756.

isabelrios:
> This should be fixed with Sasha's PR #756.

Great! I was about to comment that this is happening to me on Pixel 3 with Lockwise 1.1.2 whenever I try to refresh my logins list.
Thanks!

devinreams:
>>This should be fixed with Sasha's PR #

>I was about to comment that this is happening to me on Pixel 3 with Lockwise 1.1.2

Oh hmmm, then I think this issue is saying its still a problem since the testing was against builds that already included the change at #756.

BaronKrause:
Happens on my Samsung S8+ since today's update.

Also I cant search passwords anymore since the on screen keyboard closes almost immediately every time. Worked prior to this update.

isabelrios:
@eliserichards FYi this is also happening every time you delete a login entry. Please see video: https://youtu.be/XoXP55Vyzf4

linacambridge:
This stands out in the log:

> 2019-06-26 15:55:39.498 25887-25914/? I/sync15::sync_multiple: Syncing passwords engine!
> 2019-06-26 15:55:39.499 25887-25914/? I/sync15::sync: Syncing collection passwords
[...]

PR #900, opened 2019-09-04
Import strings from android-l10n

Pike:
State: mozilla-l10n/android-l10n@6d2fe8331a703e7f46050b93ee4db4812af36095


eliserichards:
The `sync_timeout` strings have been removed as part of #791, so they should not be re-added here @Pike

Pike:
> The `sync_timeout` strings have been removed as part of #791, so they should not be re-added here @Pike

We merged that in https://github.com/mozilla-l10n/android-l10n/pull/75/files, which isn't part of this import and will only be impacted if we force-sync from Pontoon to android-l10n.

CC @Delphine

Pike:
> Can we force sync so that we have the correct strings @Pike @Delphine?

I ended up removing them manually on the android-l10n repo, which pontoon syncs from. force-sync didn't end up doing what we had hoped for.

Pike:
Updated the PR.

PR #902, opened 2019-09-04
Add null text value for empty usernames

### LINKED ISSUE #758: Filling hidden username form field with emtpy username
Hi 👋 !

Really like using Lockwise and bumped into this oddity: I have a form that only asks for a password and Lockwise/Firefox correctly shows "no username" in the app. When I tell it to autofill the form and then press submit I get "wrong password" back from the server.

The problem turns out to be that this particular form has a hidden field called `username` that contains "something set by the server" and should be left alone. However when I inspect the request POST'ed I see that the username parameter is set to `""` (the empty string). My guess is that Lockwise fills in the form field with the "empty username" value.

I think this is a pretty weird form, so I am not sure what Lockwise should do here. Maybe checking if a form field is visible/hidden before filling it would be a way forward. Not sure I can think of a case where there are invisible (to humans) form fields that the website would expect me to fill in.

I am actually not sure if this should be a Lockwise ticket or is a more general Firefox form filling with credentials issue. If you could let me know that would be great.

App version: 1.1.1 (4053)
Firefox beta 68.0b7
Android version 9

jhugman:
Fixes #758.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/autofill/ViewNodeNavigator.kt
+++ b/app/src/main/java/mozilla/lockbox/autofill/ViewNodeNavigator.kt
@@ -5,6 +5,7 @@ import android.app.assist.AssistStructure.ViewNode
 import android.os.Build
 import android.view.autofill.AutofillId
 import androidx.annotation.RequiresApi
+import java.util.Locale
 
 interface AutofillNodeNavigator<T, U> {
     val rootNodes: List<T>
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -33,7 +33,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath 'com.android.tools.build:gradle:3.5.0'
```
jhugman: Not necessary for this PR, but upgrading as part of our normal code cycle.

jhugman:
Tested for crash with STR in #488 (fixed by #496)— no crash. Please review when ready, @eliserichards

app/src/main/java/mozilla/lockbox/autofill/ViewNodeNavigator.kt
```diff
@@ -56,13 +57,12 @@ class ViewNodeNavigator(
...
-        return node.className == "android.widget.EditText"
-    }
+    override fun isEditText(node: ViewNode) =
+        node.className == "android.widget.EditText"
 
-    override fun isHtmlInputField(node: ViewNode): Boolean {
-        return node.htmlInfo?.tag?.toLowerCase() == "input"
-    }
+    override fun isHtmlInputField(node: ViewNode) =
+        // Use English locale, as the HTML tags are all in English.
```
eliserichards: 👍

PR #909, opened 2019-09-06
603: Error states and text validation for edit view

### LINKED ISSUE #603: Error states and field validation for edit
When I go to edit an existing credential, I want to know when I’ve either missed a required field or entered incorrect data. I don’t want to save information that isn’t useful to me when I next go to log into that account.

## Requirements
* Should detect and then serve an error message when nothing OR an invalid URL is entered into the web address
* Should detect and then serve an error message [Password cannot be empty] when nothing is entered into the password field
* Error messages should provide helpful guidance for correcting the data

## Visual Design

Zeplin: https://app.zeplin.io/project/5b6895bfe4af825140aa8dbc/screen/5c548fc49c3e07307b35a574
Invision: https://mozilla.invisionapp.com/share/3SN35UHWJDX#/screens/344499813

Screen Shot 2019-07-11 at 9 26 13 AM

eliserichards:
Fixes #603 

### Engineering tasks

- [x] hide keyboard when clicking out of the edittext
- [x] show password when focused
- [x] change warning icon
- [x] change text color on warning
- [x] show warning text
- [x] warning should replace the helper text
- [x] when username is empty, make sure the name of the field stays above

### Other

- [ ] add unit tests


app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,6 +58,130 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+            ) {
+            }
+
+            override fun onTextChanged(
+                charSequence: CharSequence,
+                start: Int,
+                before: Int,
+                count: Int
+            ) {
+            }
```
jhugman: Nit: empty blocks should have a comment saying their empty. 

e.g. 

```kotlin
{
    // NOOP
}

```

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,6 +58,130 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+            override fun afterTextChanged(editable: Editable) {
+                val inputText: String? = errorLayout.editText?.text.toString()
+
+                when (errorLayout.id) {
+                    R.id.inputLayoutHostname -> {
+                        // hostname cannot be null
+                        // has to have http:// or https://
+                        when {
+                            TextUtils.isEmpty(inputText)
+                                || (!URLUtil.isHttpUrl(inputText) and !URLUtil.isHttpsUrl(inputText))
```
mnoorenberghe: We don't have such a restriction in any of the Firefox browsers so I don't see why we should have that in Lockwise. The origin field is a URI, not a URL e.g. `ftp://example.com`, `file:///`, and `chrome://FirefoxAccounts` are all values that we currently store and Sync.

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,6 +58,130 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+            ) {
+            }
+
+            override fun afterTextChanged(editable: Editable) {
+                val inputText: String? = errorLayout.editText?.text.toString()
+
+                when (errorLayout.id) {
+                    R.id.inputLayoutHostname -> {
+                        // hostname cannot be null
+                        // has to have http:// or https://
```
mnoorenberghe: You should also ensure that there are no path segments i.e. https://www.example.com is valid whereas https://www.example.com/ and https://www.example.com/foo are invalid.

mnoorenberghe: We decided not to allow editing of this field on desktop for now because users can make changes to it which break autofill so you could also de-prioritize editing of the origin and only allow editing this field for new logins though my comments here about not including path segments or limiting the scheme still applies.

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,6 +58,130 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+            if (!hasFocus) {
+                closeKeyboard()
+            }
+        }
+
+        view.inputUsername.onFocusChangeListener = View.OnFocusChangeListener { _, hasFocus ->
+            if (!hasFocus) {
+                closeKeyboard()
+            }
+        }
```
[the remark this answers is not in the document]
eliserichards: Nice, I didn't realize that you could use the same `listener` for multiple items

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,6 +58,130 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+        view.inputUsername.addTextChangedListener(
+            buildTextWatcher(
+                view.inputLayoutUsername
+            )
+        )
+        view.inputPassword.addTextChangedListener(
+            buildTextWatcher(
+                view.inputLayoutPassword
+            )
+        )
```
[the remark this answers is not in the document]
eliserichards: yes indeed!

app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
```diff
@@ -116,8 +124,8 @@ class EditItemPresenter(
 
         view.saveEntryClicks
             .subscribe {
-                credentials?.let { c ->
-                    dispatcher.dispatch(DataStoreAction.UpdateItemDetail(c))
+                credentials?.let {
+                    dispatcher.dispatch(DataStoreAction.UpdateItemDetail(it))
```
jhugman: 😻

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,48 +82,152 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         setupToolbar(view.toolbar)
-    }
+        setTextWatcher(view)
+        setKeyboardFocus(view)
```
jhugman: Nit: use `setupX(view)` instead of `setX(view)`. 

If you're setting X, this would be analagous to `this.X = view`, which is not what you're doing.

app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -49,48 +82,152 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
+                    }
+                    R.id.inputLayoutUsername -> {
+                        handleUsernameChanges(errorLayout, inputText)
+                    }
+                    R.id.inputLayoutPassword -> {
+                        handlePasswordChanges(errorLayout, inputText)
+                    }
+                    else -> {
+                        errorLayout.error = null
+                    }
```
jhugman: Unnit: This is the case I was talking about: using `switch`/`when` especially in multiple places is an indicator that you should be using a State pattern, so a class for each state. In this case, this would map to a `TextWatcher` class per input.

The trade offs for this aren't obvious, especially on constrained devices with low method count limits.

PR #922, opened 2019-09-13
(WIP) 920-add-new-edit-screenshots-tests

### LINKED ISSUE #920: [l10nScreenshotTests] Add test for Edit menu
It would be nice to have the screenshots tests updated with the new views added so that we can check that there are no translation issues.

isabelrios:
Fixes #920

Creating the PR to save the changes done but waiting for adding the missing screen (edit disclaimer view). If it is not possible to get that screen when the screenshots are needed then at least we could merge this and get all the other screenshots up to date

### DIFF
--- a/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
@@ -24,7 +24,6 @@ import org.junit.Test
 import org.junit.runner.RunWith
 
 import br.com.concretesolutions.kappuccino.custom.recyclerView.RecyclerViewInteractions.recyclerView
-
 import tools.fastlane.screengrab.Screengrab
 import tools.fastlane.screengrab.locale.LocaleTestRule
 import tools.fastlane.screengrab.UiAutomatorScreenshotStrategy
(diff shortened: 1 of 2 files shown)


isabelrios:
@eliserichards this is ready for review! I thought I have landed this but not yet 🤦‍♀
Although it is not in master I generated the screenshots, you can see all them [here](https://storage.googleapis.com/l10n-lockwise-android/2019-10-09/android/screenshots.html) (FYi @changecourse in case you are insterested)

Please let me know if there is any language missing and if you want me to open issues for some of the missing translations...
Thanks!

changecourse:
thanks for this @isabelrios, I'll coordinate app store screenshots with marketing.  

one question: does this link update itself on a somewhat regular basis, or will i need to request updates as the master branch changes over time?

isabelrios:
> one question: does this link update itself on a somewhat regular basis, or will i need to request updates as the master branch changes over time?

@changecourse No, it does not happen automatically (yet). Whenever someone wants a new set of screenshots, the script is launch manually to get them and then there is another script to upload them to gcloud. 
In gcloud we have different folders for each product and each version of screenshots is stored in a different folder with the date:
Screenshot 2019-10-09 at 19 32 58

This set up can be changed, we design it as a first approach for getting the screenshots when needed but can be modified according to other teams needs

app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
```diff
@@ -42,15 +42,15 @@ open class ScreenshotsTest {
     val localeTestRule = LocaleTestRule()
 
     @Test
-    fun testCompleteOnboarding() {
+    fun testThroughoutAllApp() {
 
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        onView(withId(R.id.buttonGetStarted))
+        onView(withId(R.id.buttonGetStartedManually))
```
jhugman: Most of these can be addressed using UI test robots. 

e.g. see `WelcomeRobot`

```kotlin
welcome {
    tapGetStarted()
}
```

isabelrios:
> While these changes work, I think I'd like to ultimately prefer to see them using the Robot pattern, which is designed for exactly this use.
> 
> We have a robot class for each of the screens, so using them, and adding to them would be a good thing to do.

@jhugman yes, I totally agree, I started this way due to some missing robots/IDs, I will create a issue to refactor all these tests to use the Robots.

PR #945, opened 2019-09-26
Cache access tokens

### LINKED ISSUE #941: Lockwise should cache access tokens
On startup (I think), `detectAccount` runs...
https://github.com/mozilla-lockwise/lockwise-android/blob/c435688538e54d309dbc94e74682c2904267363f/app/src/main/java/mozilla/lockbox/store/AccountStore.kt#L167
... and for existing users it requests a new access token.
https://github.com/mozilla-lockwise/lockwise-android/blob/c435688538e54d309dbc94e74682c2904267363f/app/src/main/java/mozilla/lockbox/store/AccountStore.kt#L197

Access tokens aren't part of the persisted FxA state. They are cached in memory by the Rust code, but in the startup scenario that cache is going to be empty.

This means that every startup of Lockwise will request a new access token, causing INSERTs in the FxA OAuth db.

We've addressed this problem in Fenix (via android-components AccountManager) by introducing a persistent caching layer (in sharedprefs) in a-c, which only requests new tokens if cached one have expired. See https://github.com/mozilla-mobile/android-components/pull/3579 for details on the problem and solution outline.

jhugman:
Fixes #941 

This PR introduces 

 * access token caching. 
 * token rotation after expiry.
 * lazy initialization of the account store during `Autofill`.

~Currently still in WIP because it hasn't passed enough smoke tests.~ Edit: fixed by `clear`ing secure preferences once the account is disconnected.

PR opened to get feedback.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -308,8 +308,7 @@ open class AccountStore(
 
         tokenRotationHandler.removeCallbacksAndMessages(null)
 
-        this.securePreferences.remove(Constant.Key.firefoxAccount)
-        this.securePreferences.remove(Constant.Key.accessToken)
+        this.securePreferences.clear()
 
         this.generateNewFirefoxAccount()
 
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -8,6 +8,7 @@ package mozilla.lockbox.store
 
 import android.content.Context
 import android.net.Uri
+import android.os.Handler
 import android.os.Looper
 import android.webkit.CookieManager
 import android.webkit.WebStorage
@@ -25,6 +26,7 @@ import kotlinx.coroutines.rx2.asSingle
 import mozilla.appservices.fxaclient.FxaException
 import mozilla.components.concept.sync.AccessTokenInfo
 import mozilla.components.concept.sync.Avatar
+import mozilla.components.concept.sync.OAuthScopedKey
 import mozilla.components.concept.sync.Profile
 import mozilla.components.service.fxa.ServerConfig
 import mozilla.components.service.fxa.FirefoxAccount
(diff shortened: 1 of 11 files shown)


jhugman:
ccing @pocmo @grigoryk r?

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -194,16 +203,56 @@ open class AccountStore(
             .subscribe(profileSubject::onNext, this::pushError)
             .addTo(compositeDisposable)
 
+        val token = securePreferences.getString(Constant.Key.accessToken)?.let {
```
grigoryk: I would instead introduce a helper class that handles token caching/rotation, and only touch a single line in the `AccountStore`. You will be able to trivially test that class's behaviour then, as well, and keep this logic simple.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
```diff
@@ -194,16 +203,56 @@ open class AccountStore(
...
+        syncCredentialSubject.onNext(credentials.asOptional())
+
+        // 2. Store this token in the secure preferences.
+        securePreferences.putString(Constant.Key.accessToken, token.toJSONObject().toString())
+
+        // 3. Schedule a token refresh. It may be quite a long time away.
+        // Calculate how long before the token expires in milliseconds.
+        val msDelay = token.expiresAt * 1000L - timingSupport.currentTimeMillis
+
+        // We'll wait until it's almost expired, leaving at most 10 minutes to fetch the token.
```
grigoryk: You don't need any of this scheduling stuff, really.

Next time you need a token, try to fetch it from your cache. If it's all good, nothing to do. If it's expired, fetch a new one from the library, cache it. That's all you need to do.

These tokens live 1-3 days, or so, so you'd only need to refresh them roughly once a day. The desire is to shorten their lifespan sometime in the future, but when that will happen is unclear, and even then it'll be on the order of a "a few hours". It's _very_ unlikely that your app won't be killed by then.

jhugman: > You don't need any of this scheduling stuff, really.
> 
> Next time you need a token, try to fetch it from your cache. If it's all good, nothing to do. If it's expired, fetch a new one from the library, cache it. That's all you need to do.

Yes this works, and would be my preference.

However, this is meant to be a small intervention. 

For right or wrong reasons, the token is derived from a-s at start up, and sent for use elsewhere. The token rotation scheduling is an unfortunate side effect of this previous decision.

Changing from "`AccountStore` gets from a-s and pushes `DataStore`" to "`DataStore` pulling from an a-s cache" model is a large amount of work. 

I would rather spend that time moving re-writing `DataStore` in terms of an `FxaManager`.

app/src/main/java/mozilla/lockbox/store/AccountStore.kt
jhugman: This is an excellent suggestion.

PR #968, opened 2019-10-11
948: prevent duplicate entries during edit

### LINKED ISSUE #948: Prevent entry duplication during edit
## Story Description

We should not be saving a duplicate of an existing login when editing an entry. It is possible to edit an entry by changing the username to conflict with one already saved for that `origin`/`formActionOrigin`/`httpRealm` combo.

### On "Edit Login" Screen

**InVision:** https://mozilla.invisionapp.com/share/QRU9LDSZ8EH
**Zeplin:** https://app.zeplin.io/project/5b6895bfe4af825140aa8dbc/screen/5d9b96331d02d5069eb3e1ba

### On "Create Login" Screen

**InVision:** https://mozilla.invisionapp.com/share/P4U9LDZ8K6N
**Zeplin:** https://app.zeplin.io/project/5b6895bfe4af825140aa8dbc/screen/5d9b962f845d361e5ec678c5

### host name

**has to have http:// or https://**
- We don't have such a restriction in any of the Firefox browsers so I don't see why we should have that in Lockwise. The origin field is a URI, not a URL e.g. ftp://example.com, file:///, and chrome://FirefoxAccounts are all values that we currently store and Sync.
- You should also ensure that there are no path segments i.e. https://www.example.com is valid whereas https://www.example.com/ and https://www.example.com/foo are invalid.
- We decided not to allow editing of this field on desktop for now because users can make changes to it which break autofill so you could also de-prioritize editing of the origin and only allow editing this field for new logins

### username
- should check (after text changes? on save?) that the username does not exist for that hostname already

eliserichards:
Fixes #948

## Testing and Review Notes

1. `Datastore` contains a list of all entries
2. `Datastore` can filter that list based on a given hostname
	- `fun getUsernamesForDomain(hostname: String): Observable<List<Optional<String>>>`
3. The edit fragment doesn’t want to directly reference the datastore (that's a job for the presenter)
4. Create a local list in the presenter that gets our list of dupes from `getUsernamesForDomain()`
5. Check this list in the fragment on text change and set error state if dupe exists

## To Do

- [x] datastore tests
- [x] edit presenter tests
- [x] remove current item from list of possible dupes
- [x] add unit tests

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
@@ -19,10 +19,10 @@ import mozilla.lockbox.extensions.filterNotNull
 import mozilla.lockbox.extensions.toDetailViewModel
 import mozilla.lockbox.flux.Dispatcher
 import mozilla.lockbox.flux.Presenter
+import mozilla.lockbox.log
 import mozilla.lockbox.model.ItemDetailViewModel
 import mozilla.lockbox.store.DataStore
 import mozilla.lockbox.store.ItemDetailStore
-import mozilla.lockbox.support.Optional
 import mozilla.lockbox.support.pushError
 
 interface EditItemDetailView {
@@ -35,7 +35,7 @@ interface EditItemDetailView {
     val hostnameChanged: Observable<CharSequence>
     val usernameChanged: Observable<CharSequence>
     val passwordChanged: Observable<CharSequence>
-    var duplicateList: List<Optional<String>>
+    var duplicateList: Set<String?>
     fun updateItem(item: ItemDetailViewModel)
     fun closeKeyboard()
 }
@@ -44,16 +44,17 @@ interface EditItemDetailView {
 class EditItemPresenter(
     private val view: EditItemDetailView,
     val itemId: String?,
+    val hostname: String,
     private val dispatcher: Dispatcher = Dispatcher.shared,
     private val dataStore: DataStore = DataStore.shared,
     private val itemDetailStore: ItemDetailStore = ItemDetailStore.shared
 ) : Presenter() {
 
     private var credentials: ServerPassword? = null
-    var listOfDuplicatesByHostname = listOf<Optional<String>>()
 
     override fun onViewReady() {
         val itemId = this.itemId ?: return
+        val hostname = this.hostname
 
         dataStore.get(itemId)
             .observeOn(mainThread())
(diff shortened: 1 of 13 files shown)


app/src/main/java/mozilla/lockbox/view/EditItemFragment.kt
```diff
@@ -233,23 +269,6 @@ class EditItemFragment : BackableFragment(), EditItemDetailView {
...
-    private fun updatePasswordVisibility(visible: Boolean) {
-        if (visible) {
-            inputPassword.transformationMethod = null
-            btnPasswordToggle.setImageResource(R.drawable.ic_hide)
-        } else {
-            inputPassword.transformationMethod = PasswordTransformationMethod.getInstance()
-            btnPasswordToggle.setImageResource(R.drawable.ic_show)
-        }
-    }
-
```
eliserichards: Moved this above with the other view `vars`

app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
```diff
@@ -62,6 +65,12 @@ class EditItemPresenter(
 
         view.isPasswordVisible = false
 
+        dataStore.getUsernamesForDomain(itemId, hostname)
+            .subscribe {
+                view.duplicateList = it
+            }
+            .addTo(compositeDisposable)
```
jhugman: This should probably be structured as to be a subscription on the `get()` above.

e.g. 

```kotlin
val getItem = dataStore.get(itemId)
    .filterNotNull()
    .doOnNext { credentials = it }

getItem
    .observeOn(mainThread())
    .subscribe(view::updateItem)
    .addTo(compositeDisposable)

getItem
    .switchMap {
        dataStore.getUsernamesForDomain(it.hostname)
    }
    .subscribe {
        view.duplicateList = it.minus(credentials.username)
    }
    .addTo(compositeDisposable)
```

Not completely happy with the use of `credentials.username` (because we're relying on a side effect of the original `getItem`).

eliserichards:
Thinking about syncing during edit & checking for duplicates. Maybe listen for `DataStoreAction.Sync` on the presenter and re-call `getUsernamesForDomain`? I'm not sure if I'm thinking along the right lines. Also thinking that maybe that would be a nice followup to this PR?

app/src/main/java/mozilla/lockbox/store/DataStore.kt
```diff
@@ -186,7 +188,19 @@ open class DataStore(
         }
     }
 
-    private fun add(item: ServerPassword) {
+    // Returns a list of usernames that match the given hostname
+    open fun getUsernamesForDomain(itemId: String, hostname: String): Observable<Set<String?>> {
+        return list.map { items ->
+            items.filter { it.hostname == hostname && it.id != itemId }
```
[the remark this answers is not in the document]
mnoorenberghe: Duplicate checking should take all 4 properties of the composite key (hostname, formSubmitURL, httpRealm, username) into account [like Toolkit](https://searchfox.org/mozilla-central/rev/17756e2a5c180d980a4b08d99f8cc0c97290ae8d/toolkit/components/passwordmgr/LoginManager.jsm#308-315). Ideally [this should be handled in a-s](https://github.com/mozilla/application-services/issues/617) so apps just need to handle the errors. On that note I would recommend renaming the function.

mnoorenberghe:
Why wasn't https://github.com/mozilla-lockwise/lockwise-android/pull/968#discussion_r335019686 addressed? This doesn't meet the requirements of deduping so this shouldn't have been merged IMO. The point of adding deduping was to be consistent with other platforms, not to make some new deduping scheme.

Issue #127, opened 2018-10-03
Use new Telemetry SDK (Glean) in Android


linuxwolf:
Update the telemetry usage from the existing `service-telemetry` component to the new Telemetry SDK once available.

linuxwolf:
Blocked waiting on the new SDK to be published

[ cc @irrationalagent to track ]

eliserichards:
@irrationalagent is this still applicable or should we close this issue?

irrationalagent:
I would say its still applicable but probably low priority

linuxwolf:
cc @georgef for additional enablement/support once we're ready here.

irrationalagent:
It is looking like lockwise may be a first customer of ecosystem telemetry. If that's the case, then the bit coming from the mobile apps will have to be implemented in glean.

PR #1070, opened 2019-11-18
Bump the required Android Components version to 19.0.1

Dexterp37:
This is the first version that comes with a Rust implementation of the Glean SDK and is the suggested starting version for integrating with the Glean SDK.

~~**Currently failing**~~

```
e: E:\Mozilla\lockwise-android\app\src\main\java\mozilla\lockbox\store\AccountStore.kt: (284, 24): Type mismatch: inferred type is KFunction1<@ParameterName String, Unit> but Consumer<in AuthFlowUrl!>! was expected
e: E:\Mozilla\lockwise-android\app\src\main\java\mozilla\lockbox\store\AccountStore.kt: (284, 60): Type mismatch: inferred type is KFunction1<@ParameterName Throwable, Unit> but Consumer<in Throwable!>! was expected

```

Connected to #127 

## To Do

- [x] fix compilation errors

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/AccountStore.kt
@@ -281,6 +281,7 @@ open class AccountStore(
 
         fxa.beginOAuthFlowAsync(Constant.FxA.scopes)
             .asMaybe(coroutineContext)
+            .map { it.url }
             .subscribe((this.loginURL as Subject)::onNext, this::pushError)
             .addTo(compositeDisposable)
     }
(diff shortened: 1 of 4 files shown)


Dexterp37:
@jhugman is this a reasonable bump? Do you have any suggestion on how to fix the build error?

Dexterp37:
> 
> 
> @jhugman is this a reasonable bump? Do you have any suggestion on how to fix the build error?

Unping, I think I fixed this :)

Dexterp37:
> 
> 
> LGTM, but I'd like a second opinion from @jhugman 👀 🤞

Totally makes sense! This is such a huge version bump that might be relevant to have QA on it, if you have resources for that!

abodea:
Verified as fixed on the `v3.3.0` with Google Pixel 3a XL(Android 10).

This being a very big update, I was not really sure what to verify. @Dexterp37 told me that I should verify that everything works as expected and Lockwise is not broken.

Note that I didn't encounter any issues(0 crashes) while performing different tests. 
Is there something else that I should look into or just the normal workflow?

I will add the `QA Verified` label based on my comment.
@eliserichards @jhugman I would appreciate a confirmation that everything looks good regarding my comment, thanks!

PR #1075, opened 2019-11-19
Fix interrupts causing the app to go to the item list.

### LINKED ISSUE #987: The delete prompt is not dismissed after a home button interrupt
Duplicate of #915 
Duplicate of #892 
Duplicate of #1016

## Steps to reproduce
1. Launch Lockwise.
2. Reach the Item Detail page.
3. Tap on the 3 dot menu and select delete.
4. Tap on the device home button in order to add Lockwise in background.
5. Resume Lockwise.
### Expected behavior
The account section is displayed with the Deleting prompt.
### Actual behavior
The entries section is displayed and the delete prompt is displayed over it.
### Device & build information
* Device: Google Pixel 3a XL(Android 9)
* Build version: Latest build from master(17/10/2019)
### Notes
Attachments: 
[Video](https://drive.google.com/file/d/1MR4AyeifFB5vLAH7g42xEzUMa_doilvs/view?usp=sharing)

jhugman:
Fixes #987

This PR makes a number of changes: 

 * stops the app re-routing to the `ItemList` in the instances of the app not being at the `LockScreen`.
 * stops the edits being overwritten by in the `EditItemDetail` screen after an interrupt; this was revealed by fixing the original bug.
 * allows dialogs to be functional after an interrupt. This was revealed by fixing the original bug.
 * allows radio dialogs after an interrupt. This is the change autolocking dialog, similar to the previous one. Again only revealed by fixing the original bug.

Note, this fixes #987 by making the app restore correctly, rather than dismissing the dialog and taking us back to itemlist.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
+++ b/app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
@@ -83,11 +83,13 @@ class EditItemPresenter(
 
         val getItem = dataStore.get(itemId)
             .filterNotNull()
+            .distinctUntilChanged()
             .doOnNext {
                 credentialsAtStart = it
             }
 
         getItem
+            .filter { credentialsToSave == null }
             .map { it.toDetailViewModel() }
             .observeOn(mainThread())
             .subscribe(view::updateItem)
--- a/app/src/main/java/mozilla/lockbox/action/RadioDialogAction.kt
+++ b/app/src/main/java/mozilla/lockbox/action/RadioDialogAction.kt
@@ -1,2 +1,0 @@
-package mozilla.lockbox.action 
-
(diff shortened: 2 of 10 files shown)


app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
```diff
@@ -83,11 +83,13 @@ class EditItemPresenter(
 
         val getItem = dataStore.get(itemId)
             .filterNotNull()
+            .distinctUntilChanged()
             .doOnNext {
                 credentialsAtStart = it
             }
 
         getItem
+            .filter { credentialsToSave == null }
```
jhugman: Don't override edits when switching apps.

PR #1080, opened 2019-11-20
Cn beta rxu

### LINKED ISSUE #1079: Add toggle button to allow users in China use local sync service
### Why/User Benefit/User Problem
Most of the Fxa users in China are now using the local service built by Beijing office. In firefox for Android and iOS we both have the toggle button to allow users to switch between local service and global service.
 
### Acceptance Criteria (how do I know when I’m done?)
If a user is now set his/her phone as locale=zh_CN & language = Chinese, and the lockwise apk is not downloaded from Google play, there should be a way for the user to switch the service before login.
460CAC24A56539AC1B0ABD839A34C4D3

2EEE4006BD5276F42AA57B1663F55EF7

rxumoz:
Fixes #1079 

## Testing and Review Notes

If a user is now set his/her phone as locale=zh_CN & language = Chinese, and the lockwise apk is not downloaded from Google play, there should be a way for the user to switch the service before login. After login, there should be no toggle button for switching service in settings.
Else there should be no "switch service" option, just like before.

## Screenshots or Videos
图片 图片

## To Do

- The existed tests are all passed but I'm still working on the tests related to these new features

- The default value of local service/global service is not really runtime. It is set during app firstrun; if change the locale and language setting later, the default value won't change with the new settings.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,8 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <AndroidXmlCodeStyleSettings>
+      <option name="ARRANGEMENT_SETTINGS_MIGRATED_TO_191" value="true" />
+    </AndroidXmlCodeStyleSettings>
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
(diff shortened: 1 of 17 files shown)


jhugman:
This is really interesting. Thank you so much for your contribution. 

I'd like to get @sandysage and @changecourse 's eyes on this.

eliserichards:
Hi @rxumoz! This is a great start, and I would love to collaborate more on this to get it ready to merge. That being said, we have limited bandwidth right now given the end-of-the-year rush. I would love to take a look at this again in January once we have some more engineering time to spare. Let me know what you think! 😄

eliserichards:
Cool! We'll make sure to work around the Spring Festival 😄 Thank you!

PR #1087, opened 2019-11-25
1086-fix-ui-tests

### LINKED ISSUE #1086: [UITests] Tests are failing after recent changes in RouteStore
Tests are failing, first test is launched but times out and the other tests do not run. See [build](https://app.bitrise.io/build/f7a90c141df45923#?tab=log)
The reason is that in commit ae6800f3315ecd34acac284d3174a00d71924c8f, the way to route to  `RouteAction.ItemList` has changed.
Tests and Navigator need to be updated so that they work after these changes

isabelrios:
Fixes #1086
PR to update and so fix the tests after recent changes in the routing

### DIFF
--- a/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
@@ -70,7 +70,7 @@ open class ScreenshotsTest {
     @Test
     fun testItemList() {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.checkAtItemList()
+        navigator.gotoItemList(false)
         Screengrab.screenshot("all-logins-screen")
 
         itemList {
@@ -90,33 +90,21 @@ open class ScreenshotsTest {
     @Test
     fun testSettingsAccountMenu() {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.checkAtItemList()
-        itemList {
-            tapAccountSetting()
-            Screengrab.screenshot("settings-account-screen")
-        }
-        accountSettingScreen { tapDisconnect()
+        navigator.gotoSettings()
+        Screengrab.screenshot("settings-menu-screen")
+        navigator.gotoAccountSetting()
+        Screengrab.screenshot("settings-account-screen")
+        accountSettingScreen {
+            tapDisconnect()
             Screengrab.screenshot("disconnect-account-screen")
             pressBack()
         }
-        pressBack()
-    }
-
-    @Test
-    fun testSettingsMenu() {
-        Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.checkAtItemList()
-        itemList {
-            tapSettings()
-            Screengrab.screenshot("settings-menu-screen")
-        }
-        pressBack()
     }
 
     @Test
     fun testEditDetailView() {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.gotoItemDetail_from_itemList(1)
+        navigator.gotoItemDetail(1)
         itemDetail {
             tapCopyPass()
             Screengrab.screenshot("password-copied-screen")
@@ -146,6 +134,5 @@ open class ScreenshotsTest {
             Screengrab.screenshot("error-empty-field")
             saveChanges()
         }
-        pressBack()
     }
 }
--- a/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
+++ b/app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
@@ -70,7 +70,7 @@ open class ScreenshotsTest {
     @Test
     fun testItemList() {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.gotoItemList(false)
+        navigator.checkAtItemList()
         Screengrab.screenshot("all-logins-screen")
 
         itemList {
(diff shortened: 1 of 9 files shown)


isabelrios:
@eliserichards @jhugman, not sure what's the best way to fix the tests. With this PR they are passing on Firebase but not locally. 
On firebase the app only starts from a clean state after each test suite, not between tests. After the changes in ae6800f, this [does not work anymore](https://github.com/mozilla-lockwise/lockwise-android/blob/master/app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt#L160):
```
fun gotoItemList(goManually: Boolean = false) {
        if (goManually) {
            bypassOnboarding()
        } else {
            Dispatcher.shared.dispatch(LifecycleAction.UseTestData)
            log.info("blocking for the routes")
            // block until onboarding
            // tap skip
            blockUntil(RouteStore.shared.routes, RouteAction.ItemList)
        }
        checkAtItemList()
    }
```
so if I get the tests running in Firebase, they don't work locally because we don't go over the Onboarding screens directly with that condition in the else.

Please let me know your thoughts as it would be nice to have the tests working again, thanks!

app/src/androidTest/java/mozilla/lockbox/screenshots/ScreenshotsTest.kt
```diff
@@ -88,24 +88,35 @@ open class ScreenshotsTest {
...
-    fun testAppSettingsMenu() {
+    fun testSettingsAccountMenu() {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
-        navigator.gotoSettings()
-        Screengrab.screenshot("settings-menu-screen")
-
-        navigator.gotoAccountSetting()
-        Screengrab.screenshot("settings-account-screen")
-
+        navigator.checkAtItemList()
```
jhugman: This could probably be replaced with by adding an `exists()` call in the `itemList` block below.

app/src/androidTest/java/mozilla/lockbox/uiTests/AppRoutePresenterTest.kt
```diff
@@ -61,11 +63,16 @@ open class AppRoutePresenterTest {
...
     }
 
     @Test
     fun testSettings() {
-        navigator.gotoSettings()
+        navigator.gotoItemList(true)
+        itemList {
+            tapSettings()
+        }
+        settings { exists() }
```
jhugman: I'm really puzzled why you're replacing these `navigator` calls with other `navigator` calls and `Robot` calls?

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
```diff
@@ -146,20 +145,16 @@ class Navigator {
...
             bypassOnboarding()
         } else {
-            Dispatcher.shared.dispatch(LifecycleAction.UseTestData)
-            log.info("blocking for the routes")
-            // block until onboarding
-            // tap skip
-            blockUntil(RouteStore.shared.routes, RouteAction.ItemList)
+            checkAtItemList()
         }
-        checkAtItemList()
```
jhugman: I think this is the root of where the problem needs to be fixed, but I don't think this will fix it.

isabelrios:
Thanks for the review and the comments @jhugman.  Hope this new solution looks better :)

isabelrios:
> The place to fix this is in `navigator.gotoItemList`, and possibly changing what `UserReset` does.

I changed that, that way I get to go to Item List without having to manually click on all onboarding options

app/src/androidTest/java/mozilla/lockbox/uiTests/AppRoutePresenterTest.kt
isabelrios: I was having issues going to the item list view. It was not possible to go there from the welcome view unless going manually through all the onboarding screens. With the latest commit here I was able to fix that. and these Robot calls have been removed

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
```diff
@@ -153,11 +152,9 @@ class Navigator {
         if (goManually) {
             bypassOnboarding()
         } else {
+            // Go to ItemList after unlocking the device
+            Dispatcher.shared.dispatch(RouteAction.LockScreen)
             Dispatcher.shared.dispatch(LifecycleAction.UseTestData)
-            log.info("blocking for the routes")
-            // block until onboarding
-            // tap skip
-            blockUntil(RouteStore.shared.routes, RouteAction.ItemList)
```
jhugman: Was this the problem?

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
```diff
@@ -153,11 +152,9 @@ class Navigator {
         if (goManually) {
             bypassOnboarding()
         } else {
+            // Go to ItemList after unlocking the device
+            Dispatcher.shared.dispatch(RouteAction.LockScreen)
```
jhugman: Was this the solution? How does this fix the problem?

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
isabelrios: Yes, the app is not able to go now from Welcome screen to  Item list automatically as it used  to do before that commit that broke the tests.

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
isabelrios: @jhugman Yes, this fixed the problem. Now the app starts in welcome screen, it goes to lock state and it can be unlocked and user is taken to Item List automatically (no need to tap on each Onboarding button). This is the only way I have found to go from Welcome screen to Item list from the tests.

app/src/androidTest/java/mozilla/lockbox/uiTests/Navigator.kt
isabelrios: @jhugman you can see that issue by running one of the tests/test suites. The app is stuck in the Welcome screen, for example run `AppRoutePresenterTest.kt`

eliserichards:
@isabelrios is this ready to be merged?

isabelrios:
> @isabelrios is this ready to be merged?

@eliserichards I was waiting for a confirmation that this solution looks fine. I have not found any other way to fix the tests. With this, they are working again, so if this looks fine, it can be merged, if not, I would need help from your (dev) side to find a better solution to have the tests running again

isabelrios:
@eliserichards, @jhugman sorry for bothering you again with this, but would like to have the UITtest working soon so that we can continue working on the manual create feature as well as the screenshots in case those are required.

Please if this solution does not look good to you, I would need some input/feedback about different ways to fix the tests, in summary how the app is re-launched after each test so that it is available in a known state. Thanks!

isabelrios:
@eliserichards @jhugman could you please take a look at this? until tests are not fixed we can't add new ones for the manual create 🙏  thanks!

PR #1123, opened 2020-01-03
824: successful create toast notification

### LINKED ISSUE #824: Display toast when new entry is saved
Depends on #821 

## Requirements
* Display toast on display item screen to confirm new entry creation
image

## Visual Design
**Zeplin:** https://zpl.io/2GRdAR3

**Icon:** [success-icon.zip](https://github.com/mozilla-lockwise/lockwise-android/files/3779773/success-icon.zip)

eliserichards:
Fixes #824 

## Testing and Review Notes
* moved the `fun showToastNotification(..)` into the `RootActivity`
* created `ToastNotificationAction`s to define each toast in a single place 

## Screenshots or Videos
save_success

## To Do

- [ ] add unit tests


app/src/main/java/mozilla/lockbox/action/RouteAction.kt
```diff
@@ -31,6 +32,12 @@ open class RouteAction(
     data class DisplayItem(val id: String) : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.entry_detail)
     data class EditItem(val id: String) : RouteAction(TelemetryEventMethod.show, TelemetryEventObject.edit_entry_detail)
 
+    data class ShowToastNotification(
+        @StringRes val strId: Int? = null,
+        val text: String? = null,
+        val viewGroup: ViewGroup
```
jhugman: Hmm.

Passing around a `ViewGroup` like this looks… weird.

I'd recommend a `sealed class` in the same way as we do for `Dialog`s: so all the `strId`s are defined in the same place.

app/src/main/java/mozilla/lockbox/view/CreateItemFragment.kt
```diff
@@ -56,4 +57,8 @@ class CreateItemFragment : ItemMutationFragment(), CreateItemView {
         view.inputUsername.hint = getString(R.string.create_username_hint_text)
         view.inputPassword.hint = getString(R.string.hint_password)
     }
-}
+
+    override fun getToastAction(strId: Int?): RouteAction {
+        return RouteAction.ShowToastNotification(strId = strId, viewGroup = this.view as ViewGroup)
```
jhugman: Removing the `viewGroup` requirement on the `Action` would make this a lot simpler.

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
```diff
@@ -51,4 +57,18 @@ class RootActivity : AppCompatActivity() {
...
+    fun showToastNotification(@StringRes strId: Int? = null, text: String? = null, viewGroup: ViewGroup) {
+        assertOnUiThread()
+        val toast = Toast(this)
+
+        toast.duration = Toast.LENGTH_SHORT
+        toast.view = layoutInflater.inflate(R.layout.toast_view, viewGroup, false)
+        toast.setGravity(Gravity.FILL_HORIZONTAL or Gravity.BOTTOM, 0, 0)
+
+        val view = toast.view.findViewById(R.id.message) as TextView
+        view.text = text?.plus(" deleted.") ?: resources.getString(strId!!)
```
jhugman: Nit: localization. 

`deleted` needs to be localizable.
`${thing} deleted` needs to be localizable. This is best done by embedding a `%1$@` (I can never remember the incantation) in the `string.xml` localization.

This will crash if neither the `text` nor the `strId` is present.

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
```diff
@@ -51,4 +57,18 @@ class RootActivity : AppCompatActivity() {
             super.onBackPressed()
         }
     }
+
+    fun showToastNotification(@StringRes strId: Int? = null, text: String? = null, viewGroup: ViewGroup) {
+        assertOnUiThread()
```
jhugman: * You should get the `ViewGroup` from the `activity_root.xml`. You can see [an example of this here][1].
 * You could feasibly move all of this into the `RoutePresenter`, because you don't need it to be callable from the fragments (again, look how we do it for `DialogAction`s.

[1]: android.widget.Toast

app/src/main/java/mozilla/lockbox/view/DisplayItemFragment.kt
```diff
@@ -192,14 +192,7 @@ class DisplayItemFragment : BackableFragment(), DisplayItemView {
...
     override fun showToastNotification(@StringRes strId: Int) {
-        assertOnUiThread()
-        val toast = Toast(activity)
-        toast.duration = Toast.LENGTH_SHORT
-        toast.view = layoutInflater.inflate(R.layout.toast_view, this.view as ViewGroup, false)
-        toast.setGravity(Gravity.FILL_HORIZONTAL or Gravity.BOTTOM, 0, 0)
-        val v = toast.view.findViewById(R.id.message) as TextView
-        v.text = resources.getString(strId)
-        toast.show()
+        (activity as RootActivity).showToastNotification(strId = strId, viewGroup = this.view as ViewGroup)
```
jhugman: This opens up another API to do the same thing. That will make things harder to change/copy/use in the future.

This should be achieved with sending a `ToastNotification` action from the presenter. Again, getting rid of the `ViewGroup` requirement makes this easier.

app/src/main/java/mozilla/lockbox/view/ItemListFragment.kt
```diff
@@ -253,28 +251,11 @@ class ItemListFragment : Fragment(), ItemListView {
     }
 
     override fun showToastNotification(@StringRes strId: Int) {
-        assertOnUiThread()
-        val toast = setUpToast(strId = strId)
-        toast.show()
+        (activity as RootActivity).showToastNotification(strId = strId, viewGroup = this.view as ViewGroup)
```
jhugman: Send `ToastAction`s instead.

app/src/main/java/mozilla/lockbox/view/ItemListFragment.kt
```diff
@@ -253,28 +251,11 @@ class ItemListFragment : Fragment(), ItemListView {
...
-
-        toast.duration = Toast.LENGTH_SHORT
-        toast.view = layoutInflater.inflate(R.layout.toast_view, this.view as ViewGroup, false)
-        toast.setGravity(Gravity.FILL_HORIZONTAL or Gravity.BOTTOM, 0, 0)
-
-        val view = toast.view.findViewById(R.id.message) as TextView
-        view.text = text?.plus(" deleted.") ?: resources.getString(strId!!)
-
-        return toast
+        (activity as RootActivity).showToastNotification(text = text, viewGroup = this.view as ViewGroup)
```
jhugman: Send `ToastAction`s instead.

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
eliserichards: "deleted" hasn't been localized.. I should reach out to Pike about that

app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
```diff
@@ -171,10 +169,11 @@ class ItemListPresenter(
             .addTo(compositeDisposable)
 
         dataStore.deletedItem
-            .subscribe {
-                val event = it.get() ?: return@subscribe
-                view.showDeleteToastNotification(event.formSubmitURL ?: event.hostname)
+            .map {
+                val event = it.get()
+                ToastNotificationAction.ShowDeleteToast(event?.formSubmitURL ?: event?.hostname)
```
jhugman: What happens when `event` is `null`?

Suggest you use 

```kolin
    .map { event ->
        event.get()?
            .let { it.formSubmitURL ?: it.hostname }
            .asOptional()
    }
    .filterNotNull()
```

app/src/main/java/mozilla/lockbox/presenter/ItemMutationPresenter.kt
```diff
@@ -146,6 +150,6 @@ abstract class ItemMutationPresenter(
     }
 
     abstract fun dismissChangesAction(hasChanges: Boolean): Action
-    abstract fun saveChangesAction(hasChanges: Boolean): ItemDetailAction
+    abstract fun saveChangesAction(hasChanges: Boolean): List<Action?>
```
jhugman: This looks as though it could be non-optional.

Also: consider changing the name to `saveChangeActions`, since this is a list now.

app/src/main/java/mozilla/lockbox/presenter/ItemMutationPresenter.kt
```diff
@@ -65,7 +65,11 @@ abstract class ItemMutationPresenter(
         view.saveEntryClicks
             .flatMap { itemDetailStore.isDirty.take(1) }
             .map { saveChangesAction(it) }
-            .subscribe(dispatcher::dispatch)
+            .subscribe { itemDetailActions ->
+                itemDetailActions.filterNotNull().forEach { action ->
+                    dispatcher.dispatch(action)
+                }
+            }
```
jhugman: If `saveChangesAction()` returns a `List<Action>`, then this could be simpler:

```kotlin
.flatMapIterable { saveChangesAction(it) }
.subscribe(dispatcher::dispatch)
```

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -94,6 +97,14 @@ abstract class RoutePresenter(
         alertDialogStore.showDialog(activity, destination)
     }
 
+    fun showToastNotification(action: ToastNotificationAction) {
+        try {
+            (activity as RootActivity).showToastNotification(action)
```
jhugman: You should inline this so the `RootPresenter`. 

It might help to start with 

```kotlin
activity.apply {
    // `this` is now the activity.
}
```

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
```diff
@@ -51,4 +58,24 @@ class RootActivity : AppCompatActivity() {
...
+
+        val toast = Toast(this)
+        toast.duration = Toast.LENGTH_SHORT
+        val container = window.decorView.rootView as ViewGroup
+        toast.view = layoutInflater.inflate(R.layout.toast_view, container, false)
+        toast.setGravity(Gravity.FILL_HORIZONTAL or Gravity.BOTTOM, 0, 0)
+
+        // set text
+        val view = toast.view.findViewById(R.id.message) as TextView
+        view.text = action.viewModel.text ?: resources.getString(action.viewModel.strId!!)
```
jhugman: I think you should change the `viewModel` here to force us to use a `strId`, with an optional `text` argument.

The we have: 

```kotlin
            // set text
            val vm = action.viewModel
            val message = toast.view.findViewById(R.id.message) as TextView
            message.text = vm.text?.let {
                resources.getString(vm.strId, it)
            } ?: resources.getString(vm.strId)
```

app/src/main/java/mozilla/lockbox/presenter/EditItemPresenter.kt
```diff
@@ -58,11 +58,11 @@ class EditItemPresenter(
     override fun saveChangesAction(hasChanges: Boolean) =
         itemId?.let {
             if (hasChanges) {
-                ItemDetailAction.EditItemSaveChanges
+                listOf<Action>(ItemDetailAction.EditItemSaveChanges)
             } else {
-                null
+                listOf(ItemDetailAction.EndEditItemSession)
```
jhugman: This can be `null`; the combination of the `let` and the `?:` means that the result will be non-null.

app/src/main/java/mozilla/lockbox/presenter/CreateItemPresenter.kt
```diff
@@ -59,7 +63,9 @@ class CreateItemPresenter(
     }
 
     override fun endEditingAction(): List<Action> {
-        return listOf(RouteAction.ItemList)
+        return listOf(
```
jhugman: This should be renamed to `endEditingActions()`

app/src/main/java/mozilla/lockbox/presenter/CreateItemPresenter.kt
```diff
@@ -42,11 +43,14 @@ class CreateItemPresenter(
             .addTo(compositeDisposable)
     }
 
-    override fun saveChangesAction(hasChanges: Boolean): ItemDetailAction {
+    override fun saveChangesAction(hasChanges: Boolean): List<Action> {
```
jhugman: Nit: rename to `saveChangeActions`.

app/src/main/java/mozilla/lockbox/action/ToastNotificationAction.kt
```diff
@@ -0,0 +1,50 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package mozilla.lockbox.action
+
+import mozilla.lockbox.R
+import mozilla.lockbox.model.ToastNotificationViewModel
+
+sealed class ToastNotificationAction(
```
jhugman: 😻

app/src/main/java/mozilla/lockbox/action/ToastNotificationAction.kt
```diff
@@ -0,0 +1,50 @@
...
+            img = R.drawable.ic_check
+        )
+    )
+
+    data class ShowDeleteToast(val text: String?) : ToastNotificationAction(
+        ToastNotificationViewModel(
+            text = text?.plus(" deleted."),
+            img = R.drawable.ic_check
+        )
+    )
```
jhugman: Nits:

 * We don't need text to be a member variable of `ShowDeleteToast`.
 * make `strId` non-optional
 * make a string resource `%1$s deleted.`

```kotlin
    data class ShowDeleteToast(text: String) : ToastNotificationAction(
        ToastNotificationViewModel(
            text = text,
            strId = R.string.thing_deleted,
            img = R.drawable.ic_check
        )
    )
```

app/src/main/java/mozilla/lockbox/model/ToastNotificationViewModel.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+package mozilla.lockbox.model
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+
+data class ToastNotificationViewModel(
+    val text: String? = null,
+    @StringRes val strId: Int? = null,
```
jhugman: Make this non-optional. The text should always be inserted into a translated string.

app/src/main/java/mozilla/lockbox/presenter/DisplayItemPresenter.kt
```diff
@@ -89,31 +87,31 @@ class DisplayItemPresenter(
...
+                dispatcher.dispatch(ToastNotificationAction.ShowCopyPasswordToast)
             }
         }
 
         handleClicks(view.passwordFieldClicks) {
             if (it.password.isNotEmpty()) {
                 dispatcher.dispatch(ClipboardAction.CopyPassword(it.password))
                 dispatcher.dispatch(DataStoreAction.Touch(it.id))
-                view.showToastNotification(R.string.toast_password_copied)
+                dispatcher.dispatch(ToastNotificationAction.ShowCopyPasswordToast)
```
jhugman: At some point, I think we're going to end up with a `vararg` version of `dispatcher.dispatch`. Is this the time to do it?

jhugman:
Also: you should also add `ToastNotificationAction` to the list of actions that need to be trimmed in the `RouteStore`. The block is [linked here][1].

[1]: https://github.com/mozilla-lockwise/lockwise-android/blob/ae6800f3315ecd34acac284d3174a00d71924c8f/app/src/main/java/mozilla/lockbox/store/RouteStore.kt#L54-L60

app/src/main/java/mozilla/lockbox/presenter/DisplayItemPresenter.kt
eliserichards: 👀

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
eliserichards: ctrl T + inline

app/src/main/java/mozilla/lockbox/presenter/DisplayItemPresenter.kt
jhugman: I think there is scope to tidy this up, but I don't think we should do it now. Let's just do the simplest thing that works right now.

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
jhugman: Note that we use `strId` in each case, but we insert the untranslatable`vm.text` in to the translatable string resource.

app/src/main/java/mozilla/lockbox/model/ToastNotificationViewModel.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package mozilla.lockbox.model
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+
+data class ToastNotificationViewModel(
+    @StringRes val strId: Int,
+    @DrawableRes val img: Int
```
jhugman: I would definitely add a

```kotlin
val untranslatableString: String? = null
```

here. Most of the time it can default to `null` but occasionally— e.g. `ShowDeleteToast` we'll need it to insert into the message.

app/src/main/res/values/strings.xml
```diff
@@ -303,4 +303,13 @@
...
     <string name="hostname_empty_invalid_text">Hostname cannot be empty</string>
 
+    <!-- This is the toast message that is shown when a new entry has been successfully created. -->
+    <string name="successful_create_toast">Login added successfully</string>
+
+    <!-- This is the content description for the icon that appears on the toast notification. -->
+    <string name="toast_icon_description">The icon that appears on a toast notification.</string>
+
+    <!-- This is the toast message that appears when an entry has been successfully deleted.  -->
+    <string name="entry_deleted_toast">%1$s deleted.</string>
```
jhugman: 😻

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
jhugman: Ah, gotcha. This _doesn't_ compile.

app/src/main/java/mozilla/lockbox/presenter/ItemListPresenter.kt
```diff
@@ -171,10 +170,11 @@ class ItemListPresenter(
...
 
         dataStore.deletedItem
-            .subscribe {
-                val event = it.get() ?: return@subscribe
-                view.showDeleteToastNotification(event.formSubmitURL ?: event.hostname)
+            .map { event ->
+                event.get()?.let { it.formSubmitURL ?: it.hostname }.asOptional()
             }
+            .filterNotNull()
+            .subscribe { dispatcher.dispatch(ToastNotificationAction.ShowDeleteToast(it)) }
```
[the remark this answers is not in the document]
eliserichards: 👍

app/src/main/java/mozilla/lockbox/view/RootActivity.kt
```diff
@@ -51,4 +58,27 @@ class RootActivity : AppCompatActivity() {
...
+            resources.getString(action.viewModel.strId)
+        } else {
+            resources.getString(action.viewModel.strId, action.viewModel.deletedItem)
+        }
+
+        // set icon
+        val image = toast.view.findViewById(R.id.icon) as ImageView
+        image.setImageResource(action.viewModel.img)
+
+        toast.show()
```
jhugman: Please inline this method where it's called.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -94,6 +98,12 @@ abstract class RoutePresenter(
         alertDialogStore.showDialog(activity, destination)
     }
 
+    fun showToastNotification(action: ToastNotificationAction) {
+        activity.apply {
+            (this as RootActivity).showToastNotification(action)
+        }
```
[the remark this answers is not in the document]
eliserichards: Should i also inline the dialog "show" method? (The thought here was to mimic what the dialog code is doing):

PR #1133, opened 2020-01-10
Fixup edit regressions introduced during edit/create refactor.

### LINKED ISSUE #1126: Error while editing Username: A login with that user name already exists error is shown
## Steps to reproduce
Tap on a login to edit it
Change the Username field 
### Expected behavior
It is  possible to edit a login
### Actual behavior
No matter the  new username you select, the error is shown: `A login with that username already exists`
### Device & build information
* Device: Pixel 2
* Build version: master 0dfe1a1
### Notes
Attachments: 
Screenshot 2020-01-07 at 12 42 55

@eliserichards this might be a regression after the manual create feature landed, since this was working fine before that.

### LINKED ISSUE #1127: Discard changes actually keeps the changes done if Save button has been pressed before
## Steps to reproduce
Open a login to edit it
Change the username value
Tap on Save button
Tap on Close button
Tap on Discard changes and Discard in the confirm dialog
### Expected behavior
Changes are not applied
### Actual behavior
Even though Discard option is selected, changes in username are not discarded

### Device & build information
* Device: Pixel 2
* Build version: master 0dfe1a1
### Notes
Attachments: Please see video: https://youtu.be/QkN9XqJVDro

jhugman:
Fixes #1126
Fixes #1127

This fixes three bugs:

 * in the edit screen, the save button saved, but did not exit the screen. This is the root cause of all these bugs.
 * in the edit screen, the duplicate usernames were updated on save, and thence caused an error to be shown. This is irrelevant if the the save button now closes the edit screen.
 * when no hostname is shown (as in create) the usernames used to detect duplicates was _all_ usernames for _all_ hostnames. This is now prevented. This was unreported.

### DIFF
--- a/app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
+++ b/app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
@@ -133,9 +133,9 @@ open class ItemDetailStore(
         itemToSave.take(1)
             .filterNotNull()
             .map { DataStoreAction.CreateItem(it) }
+            .doAfterNext { stopCreating() }
             .subscribe(dispatcher::dispatch)
             .addTo(sessionDisposable)
-        stopCreating()
     }
 
     private fun stopCreating() {
@@ -190,6 +190,7 @@ open class ItemDetailStore(
             .map { (previous, next) ->
                 DataStoreAction.UpdateItemDetail(previous, next)
             }
+            .doAfterNext { stopEditing() }
             .subscribe(dispatcher::dispatch)
             .addTo(sessionDisposable)
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/mozilla/lockbox/store/ItemDetailStore.kt
```diff
@@ -154,13 +154,14 @@ open class ItemDetailStore(
...
-        Observables.combineLatest(getItem, dataStore.list)
+        Observables.combineLatest(getItem, dataStore.list.take(1))
             .map { (opt, list) ->
                 val item = opt.value ?: return@map emptySet<String>()
 
                 val usernames = list
                     .filter(
-                        hostname = item.hostname,
+                        // Default to a blank space, otherwise we get all usernames for any hostname.
+                        hostname = if (item.hostname.isNullOrBlank()) " " else item.hostname,
```
eliserichards: Why would the `hostname` ever be null/blank? This doesn't seem like a valid case to me

jhugman: `hostname` is blank in the case of Manual Create.

PR #1142, opened 2020-01-17
Update string names to send update upstream to l10n

eliserichards:
Fixes #1138 

* Changes the name of the onboarding strings to force an update upstream in l10n (the security onboarding string was previously two different strings, but has been simplified to one - see #1138)


Pike:
Can you detail on why you changed three string IDs here?

CC @Delphine

Delphine:
@eliserichards: I think the question here is why you have changed three string IDs when only one string needed to be changed. In fact, we're going to lose existing translations for all three strings since the IDs are changed).
Is it for consistency in string IDs perhaps? Otherwise, it would be better to just change one.
thanks!

eliserichards:
Yes it is for consistency. Since I changed one of the string names it was necessary to change the two others that relate to it. @delphine

Delphine:
Ok thanks Elise!

PR #1146, opened 2020-01-23
Use red trash icon and fix deleted string

eliserichards:
image


abodea:
Verified as fixed on **v.4.0.0(6016)** using **Google Pixel 3 XL(Android 9), Samsung Galaxy S10+(Android 9).**

Issue #1219, opened 2020-09-04
Disable "null" toast


aero31aero:
## Steps to reproduce

- Set Lockwise as Autofill Provider on Android 11
- Tap on a textbox that doesn't need Autofill (I'm not sure on this part, happens on a lot of textboxes).

### Expected behavior

Nothing happens

### Actual behavior

A toast appears with text "null".

### Device & build information
* Device: Pixel 2 XL (taimen)
* Build version: 4.0.1
### Notes

Another person could reproduce it at https://www.reddit.com/r/android_beta/comments/hs0d6k/null_toast/fyqnoox

Attachments:

GlenMerlin:
Can reproduce

Pixel 4 Android 11 stable
build version 4.0.1

only happens when lockwise is the currently selected autofill app

best apps I've found to test with consistent results are 
Signal Private Messenger
Firefox Android
Instagram (Direct Messaging)

Note: the toast itself is coming from android system but due to it only happening when lockwise is the default I'm led to believe something is erroring with android 11

caeci11ius:
Confirmed occurs with lockwise as autofill provider, does not occur when provider changed to google. Problem comes back when lockwise selected again.

AMonreale:
Same thing here, Pixel 3a with Android 11. Authy seems the more consistent app showing this toast.

nsghub:
Same here on Pixel 3a and Android 11. Pretty anoying.

rickhall:
Same issue for me on my Pixel 4XL...I'll have to check on my Pixel 4a later. For me, I can easily reproduce it by going into the Messages app and entering any conversation. Everything time I enter any conversation, I get a "null" popup message.

teopb:
Happens on pixel 4a updated to android 11 today.

marcomg:
Some issue with a pixel 3a with Android 11!

oniGino:
Confirmed on pixel 2 w/android 11

blitz:
Same here with a Pixel 3a and Android 11.

@eliserichards Do you someone who can take a look?

LRomuloux:
same issue. Pixel 2 Android 11.

JoeCactus:
Replicated on Pixel 3a XL, 4.02 not yet available on Play Store for me.

blitz:
@Baron-Severin @eliserichards Thank you so much for fixing this! :rocket:

rahan-crao:
[...]

PR #1227, opened 2020-09-22
Increase version number to 4.0.2 and update release notes

eliserichards:
# Release Notes

## 4.0.2 (Build 6460)

Adds a fix for an autofill failure on Android 11.

Changes since last build:

- Fixes a 'null' toast on autofill failure on Android 11 (#1219)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,8 +22,8 @@ android {
         minSdkVersion 24
         //noinspection OldTargetApi
         targetSdkVersion 28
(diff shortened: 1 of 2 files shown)


sawatzky:
@eliserichards Sorry to hassle you, but any idea when this build will be pushed through Google Play?  Or, is there a way to sideload a signed apk from elsewhere?

This bug is driving me nuts on my Pixel...

PR #1242, opened 2020-11-18
Upgrade compile and target SDK to 29

jonalmeida:
Fixes #1241

## Testing and Review Notes

~I wasn't able to test these changes but I wrote some tests or fixed the ones that mattered.~ I did some testing but a round of QA for this with varying network connectivity would be great.

## To Do

- add “WIP” to the PR title if pushing up but not complete nor ready for review
- [x] double check the original issue to confirm it is fully satisfied
- [x] add testing notes and screenshots in PR description to help guide reviewers
- [x] add unit tests
  - optional: consider adding instrumentation (integration/UI) tests
- consider running this branch in a debug simulator and check for memory leak notifications or any warnings
- [x] request the "UX" team perform a design review (if/when applicable)
- [x] make sure CI builds are passing (e.g.: fix lint and other errors)
- [x] check on the [accessibility](https://mozilla-lockwise.github.io/lockwise-android/accessibility/) of any added UI

PR #1243, opened 2020-11-19
Upgrade compile and target SDK to 29

### LINKED ISSUE #1241: Upgrade to Android SDK 29
To be able to update Lockwise on the Google Play Store, we need to update both `compileSdkVersion` and `targetSdkVersion` to `29`. A quick try in PR #1240 shows that this results in a number of deprecation and compile errors that we need to get past first.

Please take a look at the [Contribution Guidelines](https://github.com/mozilla-lockwise/lockwise-android/blob/master/docs/contributing.md) if you are interested in helping out with this issue.

jonalmeida:
Fixes #1241

## Testing and Review Notes

~I wasn't able to test these changes but I wrote some tests or fixed the ones that mattered.~ I did some testing but a round of QA for this with varying network connectivity would be great.

## To Do

- [x] add unit tests
- [x] check on the [accessibility](https://mozilla-lockwise.github.io/lockwise-android/accessibility/) of any added UI


app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
```diff
@@ -209,6 +209,7 @@ abstract class RoutePresenter(
         }
     }
 
+    @Suppress("DEPRECATION") // Fixing the deprecation requires significant changes.
```
jhugman, line 212: Do we know when this is going to be removed? I hope we never have to find out, but would be good to add to your comment.

jhugman (approved):
I'm confused why this replaced #1242. The bulk of my review is there.

Thanks for jumping on this @jonalmeida . 

The bits to do with the issue look to be just fine.

There are 15 files out of 37 where you're changing to use `requireView()` and `requireContext()` and at least one file fixing semantically valid code of dubious quality. _At this stage_ of the release cycle, I'd encourage those changes— mechanical as they are— to be done in a separate PR.

app/src/main/java/mozilla/lockbox/presenter/RoutePresenter.kt
jonalmeida: We would only have to address if we move to a newer `targetVersion` that no longer has this API and I think we'd have enough time by then to decide at that point.

jonalmeida:
> I'm confused why this replaced #1242. The bulk of my review is there.

I believe there was some CI issues with a PR not running all the things unless it was from a branch part of the main project, so Stefan recommended a new PR that wasn't based on my fork.

> There are 15 files out of 37 where you're changing to use requireView() and requireContext() and at least one file fixing semantically valid code of dubious quality. At this stage of the release cycle, I'd encourage those changes— mechanical as they are— to be done in a separate PR.

I separated the mechanical parts in the first commit and the logic part in the second (connectivity API changes).

> fixing semantically valid code of dubious quality

It was typing fatigue really, but does not change the behaviour.

---

Merging the review comments from the old PR here:

> Should this be at 28? or at 24? or 29? Either a comment in the code or here.

This would usually match the target version implicitly, but Robolectric does not currently target 29 - we had to do this in AC as well.

I'm not sure what to do about other comments, they seem more informational to me.

